Issue #4, opened 2014-01-31
Integrate File Transfer


sualko:

itJunky:
Yes, we want it!

plberd:
I'd love that!
Specially if compatible with the way Conversations (an Android XMPP client) does them. Up until now, jsxc shows me an incoming call when Conversation users send me a file.

sualko:
I guess Conversations use [XEP-0234](https://xmpp.org/extensions/xep-0234.html) and that is the reason why you see a file transfer as an incoming call. This should be fixed with the migration to [jingle.js](https://github.com/jsxc/jsxc/commit/efcff4f0bef8ed6282bc111041740c377250569d). 

I think we also go with XEP-0234, because jingle.js implemented it already and we "only" have to make some ui adjustments and file processing. ;-)

PR #114, opened 2014-12-20
[WIP] XEP-0045 Multi-User Chat

sualko:
This is a implementation of #5.

http://xmpp.org/extensions/xep-0045.html

I opened this pull request to collect some requirements for the upcoming 2.0 release:

**Must have:**
- [x] Discover server support
- [x] Create room
- [x] Join room
- [x] Password protected rooms
- [x] Groupchat
- [x] Groupchat window design
- [x] Close room
- [x] Leave room
- [x] Handle nickname conflict

**Should have:**
- [x] Show member list
- [x] Querying for Room Information
- [x] Discover Rooms


sualko:
Currently I am thinking about the chat message design:

screen shot 2015-04-08 at 13 26 28
screen shot 2015-04-08 at 13 34 46
screen shot 2015-04-08 at 13 37 01

I prefer design 1, but what do you think?

sualko:
I am also not sure how to handle nickname and jid, because you could use <code>foo</code> as nickname, while your jid is <code>bar@localhost</code>. Should the avatar correspond to jid or nickname?

Currently I prefer to use the corresponding jid for avatar detection/generation and nickname just as alias or something like that. Maybe we could add the nickname next to the timestamp...

MarcelWaldvogel:
I would prefer design 3, if it would be possible to include the avatar somehow. 

Design 1 with mouse-over for more information and avatars – where provided – looks like the most versatile. 

How about including the (nick-)name right-aligned in the time row? That would allow the users to follow conversations even in big rooms.

shaggybb:
I'm working in something similar but in my case my XMPP only works with member_only chatrooms.

The approach:
- Each user connects to the room with Strophe.getNodeFromJid
- Member's names is the FN (Full Name) of the vcard. 
- Member list is shown after room's title with an elipsis (if necessary) and a tooltip to show all names
- There will be settings options to manage members
- Each message has his owner name.

Here some graphical examples:
group_with_disabled_input
room_style_4

When I have something more stable I will make a pull request. Now I'm fighting with older browsers like Internet Explorer 8 & 9

v1r0x:
I would prefer design 1.
In this design you can recognize the user by his avatar (in case he changed it to something other than the default).
I would also correspond the avatar to the jid (this is only the case if no avatar has been set by the user). And as @MarcelWaldvogel  suggested: A mouse over with jid, nickname and other information would be great.

sualko:
Thanks for your comments and we have a winner:

Design 1 + Tooltip + Nickname next to the timestamp

@shaggybb I see a paperclip in your last screenshot. Did you implement file transfer?

shaggybb:
@sualko really not, my code is only able to upload files to middleware (php server for instance) but not peer 2 peer. The next step will be the use of strophe.si-filetransfer.js but I have no immediate plans for it.

sualko:
Thought we could close #4 :-)...

I am facing some issues with our design:

- There are also anonymous and semi-anonymous rooms, which do not announce the users jid
- Discussion history only contains nicknames
- Same nickname does not mean same user

We could use the jid avatar if possible and otherwise the nickname avatar, but I am not sure if this more confusing as always using the nickname avatar.

sualko:
Just an update with tooltip:

screen shot 2015-04-09 at 15 55 37

MarcelWaldvogel:
I would use the pale yellow color common to tooltips instead of white

sualko:
I am thinking about to move all "could have" to release 2.1, because I want to release 2.0-beta if possible next week. This would mean, that you can only create rooms with the default server settings and could not invite contacts to a room. So what do you think?

v1r0x:
Beta sounds great!
But I don't know, what this restriction exactly means for users?

v1r0x:
Thanks for the clarification.
Sounds good! :+1:

v1r0x:
Is it possible to use the new master branch in owncloud?

sualko:
Please keep calm, I wanted to merge into develop. Now I have to check, what is going on.

v1r0x:
Ok, thanks. Didn't want to sound rude, just excited to use it ;)

sualko:
My plan was to merge first muc into develop and then develop into muc, but what ever.

I am excited to hear that you want to test muc, but currently we have no documentation for a testing environment. Please stay tuned until tomorrow, when I will hopefully release the first beta version.

PR #120, opened 2015-01-14
Display the timestamp of messages

Zauberstuhl:
- [x] fix disappearing zero
- [x]  display the date only if it's not today

message_timestamp

related to diaspora/jsxc#74


sualko:
I added support for offline messages according to [XEP-0203](http://xmpp.org/extensions/xep-0203.html) in 01c34f46dd55b537a823adb3e0b091da92ae8989.

After arlolra/otr#60 is merged, I will merge this.

Issue #118, opened 2015-01-07
Port CSS to SCSS


dimaursu:
Hi there, 
I did some work recently in porting the project's CSS files to SCSS. More details in this pull-request -> 
https://github.com/diaspora/jsxc/pull/66

Is this stuff interesting to you? 
Can you cherry-pick the changes, or should I rebase to one of your branches, and submit a pull request?

sualko:
That looks very interesting. I will try to have a deeper look at it this week.

sualko:
Your work looks great and I would be happy to see a pull request of your changes, because that should simplify many things. If you need help, I will assist.

Issue #121, opened 2015-01-18
Code organisation


sualko:
The main file has now more than 6000 loc and I feel increasingly uncomfortable with that. 

@dimaursu @Zauberstuhl  What is your though on splitting up the file in there main namespaces (e.g. xmpp, storage, ...) and to create po files for localisation? We could use online translation management systems like [transifex](https://www.transifex.com/) or [weblate](http://weblate.org/) and convert it into json with [grunt-po-json](https://www.npmjs.com/package/grunt-po-json).

@MarcelWaldvogel and I also want to move jsxc to its own organisation to simplify further development and collaboration.

Hopefully we can have a open discussion about those topics.

dimaursu:
It would be a nice idea. It could also be applied to CSS, those files are huge. I will cherry-pick my SCSS changes to a clean branch today, btw. I can split the files up right away.

Zauberstuhl:
That are some interesting points. What do you think about [i18next](http://i18next.com/)? The library supports [json files](https://github.com/diaspora/jsxc/tree/develop/locales), too.

Since diaspora translation is handled by [webtranslateit](https://webtranslateit.com/en), I created also for jsxc a [project](https://webtranslateit.com/en/projects/10365-JSXC).

I could restore the old PR for upstream if that is wanted.

sualko:
@dimaursu That is a great idea.

@Zauberstuhl I will take a deeper look into i18next, looks interesting.

sualko:
i18next solves some of my concerns about large translation files and I would appreciate if you could restore your pull request. 

Is there any change to separate jsxc from diaspora in webtranslateit? It is only for cosmetic reasons. @MarcelWaldvogel What do you think about it?

Zauberstuhl:
@jhass if @sualko is moving to webtransletit there is no need for us to have a separate translation-repo. My suggestion would be; Create a new jsxc-repo and after successful integration into jsxc we should remove the diaspora-jsxc-repo. Objections?

jhass:
I'm sure if we create a jsxc organization and then shoot Eduard a mail, he's able to move it over.

@sualko the main reason we started our own translations is that we need to replace some terminology, most importantly we talk about contacts instead of buddies, would you be okay with inheriting that change then?

sualko:
@jhass The term buddy is only used in the xmpp spec, so I have no problem with that.

I will create a new project on webtranslateit.

sualko:
[...]

PR #122, opened 2015-01-18
Conversion to SCSS

dimaursu:
#118 
The port took place initially from the diaspora/jsxc repo, there might be some differences.


sualko:
I will remove those differences and merge afterwards. Thanks for your contribution.

sualko:
Please checkout https://github.com/sualko/jsxc/compare/dimaursu:scss...sualko:dimaursu-scss

I think we should discuss two more topics:

- **variable names** Maybe we should not use color values as names, because if we change the value the name will not fit anymore.
- **generated files** I would like to put all generated files (css and js, see #121) to gitignore and use grunt-watch while developing.

dimaursu:
I think I mentioned it in the issue. Yes, the color names should be definitely more generic, or semantic, not sure how to call them. Also, there are a few that are not yet extacted. 
One more thing - for the different shades of colors, it makes sense to use in some places the Sass functions `lighten` and `darken`, like so ->  `ligthten($color, 10%)`

sualko:
I see <code>_colors</code> as a kind of config file and therefore I suggest we use functional names, like <code>$roster_bg</code> and align those variables "group color variables". e.g.

```
$color1: #xxx;

$roster_bg: $color1;
$chatwindow_bg: $color1;
```

Do you agree, or do you have a better solution?

sualko:
colors

On this image you see, which colors I combined.

I also created a jsxc and webrtc module to be able to replace the color file in other projects.

Issue #124, opened 2015-02-03
v1.1.0 beta


sualko:
Hi guys,

I just build [v1.1.0-beta](https://github.com/sualko/jsxc/releases/tag/v1.1.0-beta). Thanks to everybody, who contributed and helped to improve jsxc. If you have time, please test and report issues.

If no showstopper bug occurs, we will release 1.1.0 around Sunday. Together with this release I will also move all repositories to the new [jsxc organisation](https://github.com/jsxc) and delete the css/ folder.

@Zauberstuhl @dimaursu @jhass

Best regards,
Klaus

dimaursu:
Nice! Good idea with the css/ folder, it confused me a lot in a diff.

Zauberstuhl:
awesome ! I will have a look at my PR asap. Good job

sualko:
@shaggybb

shaggybb:
Nice! By the moment, everything seems work properly. 

I love ajax login!

shaggybb:
Hi @sualko,

I have found that jsxc.gui.init method is called twice and it duplicates the jsxc_windowList and jsxc_windowListSB divs on the body. The first call is into jsxc.init and second into jsxc.xmpp.connected.

To avoid duplication, I have done this pull request #126 to try to resolve it. May be, there is another way to do that but for the moment it avoids the problem.

PR #126, opened 2015-02-06
Avoid duplicate window list

shaggybb:
#124


sualko:
> I have found that jsxc.gui.init method is called twice and it duplicates the jsxc_windowList and jsxc_windowListSB divs on the body. The first call is into jsxc.init and second into jsxc.xmpp.connected.

You are right, this could happen if you use the new ajax login.

PR #134, opened 2015-02-19
Fixed broken grunt dataUri task

Zauberstuhl:
In `build/css/<file>` you are pointing to `../img/<file>` so we have to set the `baseDir` to the build directory otherwise grunt will not find the images! Also we have to wait for the grunt-copy-task. Before that `build` does not exist.


sualko:
Thanks for reporting. I think we don't have to fix dir level, because we use the sass engine for that.

PR #136, opened 2015-03-03
code organisation

sualko:
According to #121, I split the main file and created a new grunt task, which builds a developer version into <code>dev/</code>.

For reasons of clarity, I would also success to put all <code>jsxc.lib.*</code> files into a new folder <code>src</code>.

Are there any counterarguments?

@dimaursu @Zauberstuhl @jhass


Zauberstuhl:
Well I have to make sure that diaspora modification get not lost while merging from upstream.
Last time, I screwed with the css-to-sass thing while merging from upstream, but I guess a simple rename in git shouldn't be a problem.

sualko:
What do you mean with a simple rename?

sualko:
I did a test merge and all my worst fears came true: most modifications got lost. 

Probably you have to first merge the commit before and split by ourself before you can do an other merge.

But I see no alternative to this step...

Zauberstuhl:
Ok I will be careful thanks

PR #157, opened 2015-04-30
Update jsxc*.js <script src= location in ./example

anonimal:
Greetings, Klaus. Please forgive me if I overlooked something in the directory design, but in my experience, the currently given locations of jsxc*.js in ./example/*.html are incorrect and as a result, the examples will not work without this patch.


sualko:
Thank you, I missed that.

PR #158, opened 2015-05-01
Trivial update: no newline at end of file

anonimal:
Hello again. I'm not sure of what platform you are working on, but if it's unixy, then this pull request would be relevant (also, it gets git-diff to stop complaining). For me, the complaint had recently popped up after I had started playing with jsxc.js (using Vim). Perhaps there are also other files without newline?


sualko:
I like your pr, but I can not merge this, because it is bad practise to change code in build/. This is also obsolete with #164.

PR #159, opened 2015-05-02
Development/beta example

anonimal:
I like the showWaitAlert feature in the example's #form and think that it would help maintain continuity between both of the included examples. I know that anyone could hack their own event, but I think that this (or something like this) should be included instead of having an idle UI that waits for the backend to connect.

Also included are some very minor formatting cleanup.

PR #161, opened 2015-05-02
A hack to reset the login form after login

anonimal:
I derped on the earlier pull request, sorry about that! This is what I meant:

I'm sure this problem has a better solution, but I will leave that to the pros. As it stands, I've only tested on Firefox 31.6.0, but the form stays grey after logout.


sualko:
I think one pr for example improvements is enough. #159

anonimal:
It is. Like I said, an accident on my part, sorry.

Issue #83, opened 2014-10-03
Improved Phone Support


Cerothen:
Good Morning,

I have managed to get the software setup with our OpenFire server and everything is working as it should so for that one I would like to extend my appreciation of the well crafted software. I would like to indicate that usability from a phone is very limited for both Owncloud and the chat client. Owncloud isnt a big deal since you can always log in without chat however if a user wanted to go specifically for the chat then the access is very limited due to resolution issues.

Are there any plans to improve the UI when it comes to usage on a phone browser?

Thanks,

sualko:
@Zauberstuhl @johangm90 @jhass I would like to switch to bootstraps modal window, because it is responsive. Do you have any issues with that?

johangm90:
i think it's a good idea

sualko:
#217 adds phone support, so stay tuned for the release of v3.0.

PR #186, opened 2015-06-16
Responsive dialog

sualko:
This is the first step to get jsxc responsive (#83).

I used magnific popup as lightbox, because it has a better api and is easier to namespace than bootstraps dialog.

One more big step was to extract all templates to single files and to include some parts of bootstrap.

After some more testing, I will merge this end of the week. Any doubts?

@johangm90 @Zauberstuhl @jhass @dimaursu 

screen shot 2015-06-16 at 10 18 56
screen shot 2015-06-16 at 10 22 51

I have just seen, that I missed to update the settings dialog.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,3 +7,9 @@
 [submodule "lib/i18next"]
 	path = lib/i18next
 	url = https://github.com/i18next/i18next.git
+[submodule "lib/bootstrap"]
+	path = lib/bootstrap
+	url = https://github.com/twbs/bootstrap-sass/
+[submodule "lib/magnific-popup"]
+	path = lib/magnific-popup
+	url = https://github.com/dimsemenov/Magnific-Popup
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ archives/
 /css/
 dev/
 .wti
+/tmp/
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -6,6 +6,8 @@ module.exports = function(grunt) {
       return '<%= target %>/' + el.file;
    });
 
+   var git_cached = [];
+
    // Project configuration.
    grunt.initConfig({
       app: grunt.file.readJSON('package.json'),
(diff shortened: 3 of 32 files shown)


.gitmodules
```diff
@@ -7,3 +7,9 @@
 [submodule "lib/i18next"]
 	path = lib/i18next
 	url = https://github.com/i18next/i18next.git
+[submodule "lib/bootstrap"]
+	path = lib/bootstrap
+	url = https://github.com/twbs/bootstrap-sass/
+[submodule "lib/magnific-popup"]
+	path = lib/magnific-popup
+	url = https://github.com/dimsemenov/Magnific-Popup
```
dimaursu: While using submodules in git is a working solution to dependency management, I don't think it's the right one. Technically speaking, these aren't submodules, as in separated parts of this project, but dependencies - we'd be better off using bower for dependency resolution. It was created for this sort of stuff, and many people use it.

sualko: Why do you think this is not the right solution? 

I know that currently many people use bower and I also used it before, but in this scenario I see no advantage. Only one disadvantage: Bower does not create a git repository. If I have to change something in strophe.jingle or otr, with git submodule I can easily commit those changes.

dimaursu: I agree it's easier to make necessary changes to those projects, but this also comes at a cost - there is a divergence growing between your version of OTR, for example, and the upstream one. If more and more patches are not included in the main distribution (even if most are accepted, let's suppose), the burden of maintaining those changes will grow larger and larger.

Other than these comments above, everything looks fine. Good job!

sualko: Starting with 7c3e4119ddeb89b822f6c73d2c9fd38281ac32ac I decided to follow your proposal , because I saw how many libs are laying around in the lib folder. So new libs will now be added through bower and old libs will be converted to bower after 2.1 was released. But this will not cause that we always use the original project, because not every project accepts pull requests (or  accepts pull requests in a reasonable time).

PR #200, opened 2015-07-30
Remove maxBytes restriction from dataUri task

Zauberstuhl:
Currently there are multiple files which are larger then the current limit.


sualko:
I agree with you, that we should include more images into our stylesheet, but I think this is the wrong way. If you put every image into the stylesheet file it will grow really fast and this will result in longer loading times and therefore a delay in the dom ready event.

We should try to minimize our images with inkscape (e.g. plain-svg) or try [grunt-svgmin](https://github.com/sindresorhus/grunt-svgmin).

But this will not go into the 2.1.0 release, which is planed for today.

Zauberstuhl:
Ok shrink it instead of increasing the limit sounds good to me too

Issue #77, opened 2014-09-23
Roster groups are ignored


ssams:
Buddies on the roster are displayed in a single list, group associations are ignored. Displaying the groups would help improving usability, especially when there are many buddies on the roster.
##

pedrobastosz:
Authors of this library, you saved me a lot of headaches with this amazing library :).

+1 for this thread, I would love it too.

bsv144:
Is the ability to group contacts in roster?

cirovargas:
OK, this is possible?

sualko:
It is possible, but not implemented. Pull request's are always welcome.

cirovargas:
Hi @sualko, where i can do this? You can help me to do faster?

I need to implement http://www.xmpp.org/extensions/xep-0083.html too

sualko:
@cirovargas please read our [developer notes](https://github.com/jsxc/jsxc/wiki/Developer-notes) and open an empty pull request to discuss your additions before you start to develop.

rasos:
+1 to see which groups are available when subscribing and list where I am already a member of.

PR #210, opened 2015-08-19
Jsxc does not distinguish between groups

Zauberstuhl:
related to diaspora/jsxc#89
related to diaspora/jsxc#117

we have duplicates in our contact list cause of the same user in multiple groups.
untitled

do you want this change in upstream too or do you plan on implementing group categorization in future?

### DIFF
--- a/src/jsxc.lib.xmpp.js
+++ b/src/jsxc.lib.xmpp.js
@@ -490,6 +490,10 @@ jsxc.xmpp = {
          var bid = jsxc.jidToBid(jid);
          var sub = $(this).attr('subscription');
 
+         // jsxc does not distinguish between groups
+         // therefore we should check on duplicates
+         if ($.inArray(bid, buddies) != -1) { return; }
+
          buddies.push(bid);
 
          jsxc.storage.removeUserItem('res', bid);


sualko:
> do you want this change in upstream too

I always want such stuff in upstream, but before I merge I want to resolve the reason for the bug. I do not think that this is group related, because the spec does not use multiple item elements for groups:

> Each <item/> element MAY contain one or more <group/> child elements, for use in collecting roster items into various categories.

http://xmpp.org/rfcs/rfc3921.html#rfc.section.7.1

Maybe you can post your roster as xml. I think also that diaspora/jsxc#89 is not related, because the onRoster function should be only called once and not on every page refresh.

> do you plan on implementing group categorization in future?

Here you go #77.

Zauberstuhl:
I should have checked the rfc first. Thanks it was maybe too late yesterday ;)

sualko:
So does this fix your issue, or not?

Zauberstuhl:
diaspora/vines#61 fixes the issue. the sql adapter created the duplicates not jsxc. Cheers!

sualko:
Great news, that you could solve your issue.

Issue #137, opened 2015-03-09
Daltonian mode ?


AugierLe42e:
This is something that has been requested by diaspora* users. JSXC currently use blue, gray and red colors which are not easy to distinguish by daltonian users. I thought about adding it in diaspora's version of JSXC but it might benefit to upstream too.
Are you interrested by this feature ?

sualko:
It depends on its realization, but in general we are interested in such topics.

Maybe you can write some lines about your ideas.

AugierLe42e:
Idea would have an optin mode for daltonian that would override the CSS rules of colors for connected, disconnected and busy status.

sualko:
Sounds great. Would love to see a PR.

AugierLe42e:
I can see what I can do.

sualko:
#217 should also satisfy daltonian.

![screen shot 2015-11-18 at 14 43 08](https://cloud.githubusercontent.com/assets/2974196/11242551/bf8c736e-8e02-11e5-8d38-24fd3c13baaa.png)
 
What do you think?

Issue #133, opened 2015-02-18
Switch from JS to CSS based animation


dimaursu:
I've implemented a while ago the animation in pure CSS, simplifying the JS code, and making the animation smoother (about a 3x gain in FPS). 
Are you interested in this? The whole trick involves a checkbox, it's quite simple. 

It might be a bit less flexible than the JS animation, it's a bit harder to get the drawer state (although not that hard, just reading the checkbox value)

https://github.com/diaspora/jsxc/commit/0fb67675e1f67464a0aad2c9054fd3d2eaae8163

sualko:
I think the commit is not enough, there should be more from diaspora/jsxc#63, but that is just an aside.

Are there any disadvantages of this method? Browser support or something like that? I never thought about css performance, but I'm always interested in learning new things (could be useful for #83).

While googling I found http://css-tricks.com/myth-busting-css-animations-vs-javascript/ and therefore https://greensock.com/jquery-gsap-plugin. Do you know gsap? That could boost every animation.

sualko:
> performance improvements are at least 3x, in Firefox benchmarks

How did you perform your benchmarks?

dimaursu:
Chromium and Firefox inspector. The "Timeline" feature shows the FPS and where was the time spent. it was 5 FPS vs 15 FPS during animation (which is still slow, quite far from 60 FPS), but definitely better. The animation was visibly smoother, mainly because of using css `translate` instead of animation on the right property. CSS translations have GPU acceleration, and do not stutter like JS ones.

dimaursu:
http://jankfree.org/ - here are the resources I've learned this stuff. Using greensock to animate a drawer - that's kinda overdoing it, it's not worth it. In terms of browser support, we're pretty much covered: 
http://caniuse.com/#search=translate
this being used for animation (but of course we can provide a css fallback with animation on the right property)
http://caniuse.com/#search=sibling
and a sibbling selector is used to adjust the stiling in relationship with the checkbox, which keeps the state of the drawer. 

One thing that could be a problem - you can't have swipes on a mobile device. That can only be achieved  with JS.

sualko:
Together with #229 and #217 this should be fixed.

PR #217, opened 2015-09-04
[WIP] New design

sualko:
This pr is to inform you about the current state of our new design.

screenshot2

- [x] remove redundant elements
- [x] update unread
- [x] update last msg
- [x] roster bottom
- [x] window design

It is currently not ready to test it with owncloud. 

@LEDfan @jancborchardt


sualko:
screenshot3

@Zauberstuhl How do you like the new design? Do you think this (in dark) will also fit in Diaspora?

Zauberstuhl:
@sualko we moved to bootstrap3 so I am pretty sure that will fit. I like the new design :+1:

sualko:
screen shot 2015-09-08 at 10 33 56
screen shot 2015-09-08 at 12 32 14

I am thinking about to make the roster wider on big screens. Any opinions?

The next big step is the window design. I am not sure, if light colours work here as well. I do not like the border, but otherwise it is hard to distinguish between window and background. A shadow would help, but in this case I think the roster needs one as well.

screen shot 2015-09-08 at 13 22 30
screen shot 2015-09-08 at 13 27 41

jancborchardt:
@sualko the roster is wide enough it seems. :) I wouldn’t necessarily make it wider on big screens.

One thing: Can you always test with latest master? Then you get a better feeling of how it will look with the latest version. For example, the avatar placeholders should not use bold letters but light ones. Also, please submit a pull request to the avatar placeholder algorithm with your color fixes. :)

jancborchardt:
Otherwise really nice, with the online dot, the color, spacing etc! :) 

The message box should not have a shadow as it looks strange and works just fine without.
* the individual messages don’t need a border
* the name in the header of the message box can be vertically centered with the avatar, as there is no second line with the latest message
* again – the message input field should be left-aligned instead of right. ;)

sualko:
Designing stuff is really annoying... So many ways to do it....

screen shot 2015-09-17 at 16 25 49

screen shot 2015-09-17 at 16 26 11

jancborchardt:
@sualko nice! The message window of the top one is better (with the light background) but combined with the message window header of the bottom one (also the lighter one).

hwdfld:
Looks quite good :) I would suggest minimizing the header in the chat window by cutting of the user icon at the bottom (fading it out). And offline users should be hidden by default. Instead of showing "Online" at the bottom right for me as a user, maybe one should replace it with my name (no the best solution I know, because my name is already been shown at the top right, but more consistent with the buddy list layout). Btw showing "Ilias" ... does it mean there is some integration going on? That would be awesome :D

sualko:
@hwdfl It is a little bit neglected and currently only for the 4.x branch, but here you go: https://github.com/jsxc/jsxc.ilias. Maybe you are also interested in our sogo version: https://github.com/jsxc/jsxc.sogo

sualko:
Related #137, #133

sualko:
@jancborchardt I knew that you would say that, but a little survey showed, that user do not like if everything is white without any contrast (true for jsxc and owncloud btw). I also think that the dark bar make sense, because we have to delimit to three directions. @LEDfan What do you think?

screen shot 2015-09-24 at 15 21 14

Of course the emoticons and presence menu are normally folded. 

@hwdfld I also tried to minimize the header, but it looks weird.

sualko:
I think we getting closer to a merging state. The only thing missing is some clean up and testing. Of course there are other tasks, like reducing dialogs, login procedure, ... but the new design should be basically done.

Desktop:
screen shot 2015-10-07 at 14 16 11

Mobile:
tmp_14227-screenshot_2015-10-07-14-16-371424115086

MarcelWaldvogel:
Looks great. The hangup overlay in the mobile environment would cover head or eyes of the peer. Can this be changed?

sualko:
You can toggle the controls with a touch on the video. I think this is 
common and should be fine for most users.

sualko:
This needs exhaustive testing, but I think it is good to merge.

jancborchardt:
Great stuff! :)

Similar as @MarcelWaldvogel mentioned, I would say that the controls should be displayed on the bottom, centered. That would avoid any possible overlap as much as possible and also move the controls close to your own video feed.

Zauberstuhl:
was pushed to latest diaspora dev-version. good job!

PR #247, opened 2015-11-22
Use crendential when performing turn credential ajax crossdomain call.

nitmir:
To be able to work cross-domain, the destination domain need CORS to be
properly configured. withCredentials means that cookie are send then
performing the ajax call, allowing all kind of cookie based authentication.


sualko:
I feel uncomfortable to send credentials in default configuration, but I also see the need for this. Can you get this configurable? As option for the <code>jsxc.options.RTCPeerConfig</code> object?

nitmir:
like this ?

PR #249, opened 2015-11-26
Fix invalid parent selector

Zauberstuhl:
will throw an error if you use a libsass >= 3.3.2

Not sure if there is a better solution for this but the current will compile into:

`#jsxc_roster > .jsxc_bottom`

which is the intended one I guess

### DIFF
--- a/scss/_roster.scss
+++ b/scss/_roster.scss
@@ -233,7 +233,7 @@
     }
 }
 
-#jsxc_roster > {
+#jsxc_roster > & {
     &.jsxc_bottom {
         position: absolute;
         left: 0;


scss/_roster.scss
```diff
@@ -233,7 +233,7 @@
     }
 }
 
-#jsxc_roster > {
+#jsxc_roster > & {
     &.jsxc_bottom {
```
sualko: I think we can remove the ampersand anyway. Should work completely without it.

Zauberstuhl: you talk about :
```
#jsxc_roster > {
     .jsxc_bottom {
```
?

that results in:
`Invalid parent selector` -> ` #jsxc_roster > { `

btw I am using:
```
node-sass 3.4.2
libsass 3.3.2
```

sualko: With node-sass@3.3.2 it works like a charm. Forcing it (#246) should fix the issue.

sualko:
I close this issue, because I want to get rid of the develop branch with the release of 3.0-beta1 today. But I keep this in mind and will update sass asap.

PR #268, opened 2016-01-22
first set of new emoticons

sualko:
Before I create them all, I would like to hear your meanings about  the new emoticons.

@Zauberstuhl @LEDfan @jancborchardt What do you think?

screenshot from 2016-01-22 13-21-51

screen shot 2016-01-22 at 13 30 29

Even if you do not like this one, we should definitely consider to replace the old ones, because there are only png and looking terrible on high-resolution devices.

### DIFF
--- a/img/emotions/angry.svg
+++ b/img/emotions/angry.svg
@@ -1,0 +1,181 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   width="50"
+   height="50"
+   id="svg2"
+   version="1.1"
+   inkscape:version="0.91+devel r"
+   sodipodi:docname="angry.svg"
+   viewBox="0 0 50 50">
+  <defs
+     id="defs4" />
+  <sodipodi:namedview
+     id="base"
+     pagecolor="#ffffff"
+     bordercolor="#666666"
+     borderopacity="1.0"
+     inkscape:pageopacity="0.0"
+     inkscape:pageshadow="2"
+     inkscape:zoom="12.996589"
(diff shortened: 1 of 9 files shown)


jancborchardt:
We should probably use some open-source set of standard emojis, no? Creating them all ourselves is a bit crazy and is not something we should really spend time on

sualko:
Do you have some? I found no good ones.

jancborchardt:
There’s two good sets:
- [Emoji One](http://emojione.com/) ([source](https://github.com/Ranks/emojione)). CC-BY licensed, and also used by Discourse.
- [Twitter’s emoji set](https://twitter.github.io/twemoji/) ([source](https://github.com/twitter/twemoji)). Also CC-BY, and used by Wordpress it seems.

Both look great and fit our style. What do you think @sualko @MorrisJobke @Henni @colmoneill?

LEDfan:
First of all they look already pretty good. (Except the half gradient :P).

You also got the Github emoji's but at the time I wanted to integrate them in the old Chat the license was very unclear about the images itself. https://github.com/github/gemoji

LEDfan:
There is also https://github.com/Genshin/PhantomOpenEmoji

LEDfan:
There are also the emoji's of Telegram https://github.com/zhukov/webogram/blob/master/app/img/emojisprite_0.png I'm not sure if we can use them.

BTW maybe it's a good idea to sprite the images? There are some Grunt plugins which you feed a list of images and they generate a sprite of it and a CSS file with the correct positions.

Henni:
A comparison of some of the most used emoji sets: http://apps.timwhitlock.info/emoji/tables/unicode

Most messengers use the apple emoji set which is also used by Github and integrated in https://github.com/github/gemoji
We could probably use this emoji set as well, the license seems fine to me.

But why not choose an open source project like http://emojione.com/. As @jancborchardt mentioned it's completely free. Only attribution is required.

jancborchardt:
Yeah, I already compared of course. PhantomOpenEmoji as well as Apple's ones are ugly as hell in my opinion. Emoji One or Twitter's are my favorite.

DanScharon:
> Yeah, I already compared of course. PhantomOpenEmoji as well as Apple's ones are ugly as hell in my opinion. Emoji One or Twitter's are my favorite.

yes, I fully agree. I would tend just a little more towards the Emoji One set.

MorrisJobke:
> yes, I fully agree. I would tend just a little more towards the Emoji One set.

Same here

jancborchardt:
Cool, let’s go for Emoji One then. @sualko @LEDfan cool?

sualko:
Ding, ding, ding. We have a winner: [emoji one](http://emojione.com).

It's also my favorite set, because it is good looking, has a great license and a javascript api. 

Thank you guys for your opinions and information. I love open source.

Zauberstuhl:
I already pushed some emoji to the diaspora repo but emojione looks better to me (especially the license). Would be nice to have! Thank u

PR #283, opened 2016-02-26
Stop each MediaStreamTrack individually

Perlence:
Method MediaStream.stop has been [deprecated](https://developers.google.com/web/updates/2015/07/mediastream-deprecations#stop-ended-and-active) since Google Chrome 45 in favor of stopping each track individually.

### DIFF
--- a/src/jsxc.lib.webrtc.js
+++ b/src/jsxc.lib.webrtc.js
@@ -620,7 +620,14 @@ jsxc.webrtc = {
       var bid = jsxc.jidToBid(session.peerID);
 
       if (this.localStream) {
-         this.localStream.stop();
+         if (typeof this.localStream.stop === 'function') {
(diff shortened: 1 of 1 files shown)


sualko:
We should be backward compatible. Could you add a check if <code>localStream.stop</code> is present? If you have time, please also squash your commits.

Perlence:
Yes, sure, I'll do it.

Perlence:
@sualko Done, checked and squashed.

sualko:
Thank you for your contribution. I will merge it, but maybe I have to change it a little bit, because I just read:

> Removing stop() from MediaStream did not remove any real functionality: processes for detaching source devices and so on have to be done on MediaStreamTrack anyway.

This is maybe the reason for #290.

PR #325, opened 2016-04-14
Fix auto-accept

Perlence:
Here's bug: auto-accept does not start the video/audio stream, i.e. it opens call window, but no multimedia data is sent.


sualko:
Thank you for your work and patience.

PR #351, opened 2016-07-28
Add support for Http File Upload (XEP-0363)

sngyai:
Add Support for Http File Upload(XEP-0363). It is needed to modify your server to enable cross-domain.


sualko:
Hey @sngyai. You opened two pull requests (#351, #352) with the same title. Can you close one of them?

PR #361, opened 2016-08-17
99% portuguese translate

hebertreis:


sualko:
Thank you for your pr. Next time it is maybe easier to use the [online translation tool](https://webtranslateit.com/en/projects/10365-JSXC/project_locales), but this is totally up to you :wink:.

PR #422, opened 2016-12-13
Multiline message

sualko:
@ChristianTacke @thfree please test and verify if possible

screenshot from 2016-12-13 13-45-22

Line breaks are possible with shift+enter.

- [x] When starting a new chat by clicking on a buddy from the buddy list, the textarea is "large".
- [x] Test mobil

### DIFF
--- a/src/jsxc.lib.gui.js
+++ b/src/jsxc.lib.gui.js
@@ -2094,10 +2094,11 @@ jsxc.gui.window = {
          return false;
       });
 
+      var textinputBlurTimeout;
       win.find('.jsxc_textinput').keyup(function(ev) {
          var body = $(this).val();
 
-         if (ev.which === 13) {
+         if (ev.which === 13 && !ev.shiftKey) {
             body = '';
          }
 
@@ -2107,7 +2108,8 @@ jsxc.gui.window = {
             jsxc.gui.window.close(bid);
          }
       }).keypress(function(ev) {
-         if (ev.which !== 13 || !$(this).val()) {
+         if (ev.which !== 13 || ev.shiftKey || !$(this).val()) {
+            resizeTextarea.call(this);
             return;
          }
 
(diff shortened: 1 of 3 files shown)


ChristianTacke:
I hope to patch my version and let people test it the next days.

template/chatWindow.html
```diff
@@ -51,7 +51,7 @@
             </div>
          </div>
          <div class="jsxc_transfer jsxc_otr jsxc_disabled" />
-         <input type="text" class="jsxc_textinput" data-i18n="[placeholder]Message" />
+         <textarea class="jsxc_textinput" data-i18n="[placeholder]Message"></textarea>
```
ChristianTacke: Why a full `</textarea>` instead of just a …` />`?

src/jsxc.lib.gui.js
```diff
@@ -2117,16 +2119,39 @@ jsxc.gui.window = {
...
 
+      function resizeTextarea() {
+         if (!$(this).data('originalHeight')) {
+            $(this).data('originalHeight', $(this).outerHeight());
+         }
+         if ($(this).outerHeight() < this.scrollHeight) {
+            $(this).height($(this).data('originalHeight') * 1.5);
+         }
+      }
+
```
ChristianTacke: Is this a good place for a helper function? In between two `win.find()` setup calls?

ChristianTacke:
For testing I applied the changes to 3.0.1, I hope the changes do not require any intermediate updates?

In testing, I noticed one interesting behaviour on Chromium 55.0.2883.75:
When starting a new chat by clicking on a buddy from the buddy list, the textarea is "large". When clicking anywhere in the whole window (even outside any jsxc content), the textarea shrinks to "small".

Also:
So it is intended, that there are two sizes? small and large? It's not intended to grow further for even more input?

template/chatWindow.html
sualko: Textarea is not allowed to be self closing, because it can contain text. Probably jquery would take care of this, but I am used to it.

src/jsxc.lib.gui.js
sualko: This is a subject for debate :laughing:, but it is absolute valid javascript. As you can see at the top of the outer function, the complete function needs to be rewritten. The reason why I defined this helper function at this place was, that it is not used elsewhere in the code and therefore it is easier to read.

sualko:
>  I hope the changes do not require any intermediate updates?

Nope.

> When starting a new chat by clicking on a buddy from the buddy list, the textarea is "large".

That is not the intended behavior.

> So it is intended, that there are two sizes? small and large? It's not intended to grow further for even more input?

Yes this is intended. I is similar to the behavior of mobil chat clients like signal. If you need more space you can enlarge the complete chat window, so that the textarea has more width. 

One thing which I didn't check it the behavior on small devices.

Thank you for testing this new feature.

src/jsxc.lib.gui.js
ChristianTacke: Okay. Maybe this would be a good start to slowly rewrite it now?

ChristianTacke:
> That is not the intended behavior.
Okay, 

One further note: I have to do the following steps to reproduce it:
* Start chat, login (using showloginbox)
* Click on one buddy (input area is normal)
* Close buddy window
* Press F5 (restart jsxc)
* Open buddy again. Now the input area is large.

> Yes this is intended. I is similar to the behavior of mobil chat clients like signal. If you need more space you can enlarge the complete chat window, so that the textarea has more width.

Just as a note: The mobile app Conversations increases the input area line by line, as you type. It increases it to the point, where your old chat is completely hidden and all your screen space is used for the input area.

src/jsxc.lib.gui.js
sualko: 4.0 will transfer jsxc to typescript, so I would like to wait after 3.1 with bigger changes :wink: .

sualko:
What do you think, can we merge it?

ChristianTacke:
Yes, go ahead.

PR #425, opened 2016-12-16
Clear userlist before appending entries from "getUsers" result.

fancycode:
If the server is returning entries slower than the user is entering
additional characters, the callback gets executed multiple times after,
the list has been cleared - resulting in the same entries being appended
multiple times.


sualko:
Thank you for your contribution. This should work in most of the cases.

PR #424, opened 2016-12-16
New feature: notification of a message composing in a chat/groupchat window

mmoqui:
Add the Strophejs plugin 'chatstates' to implement a new functionality: to be informed in the chat or the groupchat window when a buddy is composing a message.

fix #9

PR #431, opened 2017-01-05
Chat State Notifications (XEP-0085)

sualko:
Improve  #424

@mmoqui Do you want to test this?

**TODO**
- [x] Add general option
- [x] Add user option


sualko:
> Ok for testing it

Great. Can you report your results? Because if you find no issues, I will merge this.

>  you've forgotten to include the chatstate Stophe plugin

I don't think so, because it also revert the revert of your pr. On the command line I see your commit and there should be also the plugin entry. We will see how this will work :wink:

sualko:
> We will see how this will work 

It will not work :laughing:. I manually have to add your changes.

src/jsxc.lib.xmpp.chatState.js
```diff
@@ -0,0 +1,247 @@
...
+   if (usersComposing.indexOf(user) >= 0) {
+      // remove user from list
+      usersComposing.splice(usersComposing.indexOf(user), 1);
+      win.data('composing', usersComposing);
+   }
+
+   if (usersComposing.length === 0) {
+      var duration = parseFloat(el.css('transition-duration'));
+
+      if (el.css('transition-duration').match(/s$/)) {
```
mmoqui: Just for information, I detected also a bug with the translations: the $t.() function doesn't take into account the variables and then we can see the name of the variable (__name__ for example) when a translated text is displayed (like in the desktop notifications)

Gruntfile.js
```diff
@@ -26,7 +26,7 @@ module.exports = function(grunt) {
...
-               src: ['lib/i18next/i18next.min.js', 'lib/jquery-i18next/jquery-i18next.min.js', 'lib/magnific-popup/dist/*.js', 'lib/favico.js/favico.js', 'lib/emojione/lib/js/*.js', 'lib/emojione/assets/svg/*.svg', 'lib/strophe.js/strophe.js', 'lib/strophe.x/*.js', 'lib/strophe.bookmarks/*.js', 'lib/strophe.vcard/*.js', 'lib/strophe.jinglejs/*-bundle.js', 'lib/otr/build/**', 'lib/otr/lib/dsa-webworker.js', 'lib/otr/lib/sm-webworker.js', 'lib/otr/lib/const.js', 'lib/otr/lib/helpers.js', 'lib/otr/lib/dsa.js', 'lib/otr/vendor/*.js', 'lib/*.js', 'LICENSE', 'img/**', 'sound/**'],
+               src: ['lib/i18next/i18next.min.js', 'lib/jquery-i18next/jquery-i18next.min.js', 'lib/magnific-popup/dist/*.js', 'lib/favico.js/favico.js', 'lib/emojione/lib/js/*.js', 'lib/emojione/assets/svg/*.svg', 'lib/strophe.js/strophe.js', 'lib/strophe.x/*.js', 'lib/strophe.bookmarks/*.js', 'lib/strophe.chatstates/*.js', 'lib/strophe.vcard/*.js', 'lib/strophe.jinglejs/*-bundle.js', 'lib/otr/build/**', 'lib/otr/lib/dsa-webworker.js', 'lib/otr/lib/sm-webworker.js', 'lib/otr/lib/const.js', 'lib/otr/lib/helpers.js', 'lib/otr/lib/dsa.js', 'lib/otr/vendor/*.js', 'lib/*.js', 'LICENSE', 'img/**', 'sound/**'],
```
ChristianTacke: Hi,
could you consider wrapping this over multiple lines? This would improve readability of the diff/patch in the future.

template/settings.html
```diff
@@ -117,3 +117,24 @@ <h3 data-i18n="Carbon_copy"></h3>
...
+            </div>
+         </div>
+      </div>
+      <div class="form-group">
+         <div class="col-sm-12">
+            <button class="btn btn-primary jsxc_continue" type="submit" data-i18n="Save"></button>
+         </div>
+      </div>
+   </fieldset>
+</form>
```
ChristianTacke: I haven't yet seen, how this will look like, but the current (old) settings window is already taking the whole screen on a 1024x1024 browser window. Are jquery ui tabs an option here?

Gruntfile.js
sualko: Good point. I will try if this is possible with `grunt-jsbeautifier`.

src/jsxc.lib.xmpp.chatState.js
sualko: I will have a look into the duration issue. Could you open another issue for the translation problem?

template/settings.html
sualko: We definitely have to improve the settings dialog, but I think this is worthy a second pr :wink:

src/jsxc.lib.xmpp.chatState.js
```diff
@@ -130,9 +130,10 @@ jsxc.xmpp.chatState.onPaused = function(ev, jid) {
    }
 
    if (usersComposing.length === 0) {
-      var duration = parseFloat(el.css('transition-duration'));
+      var durationValue = el.css('transition-duration') || '0';
```
ChristianTacke: Why not ` || '0s'`? I think that's the more or less offiecial default, according to https://developer.mozilla.org/en-US/docs/Web/CSS/transition-duration#Summary

src/jsxc.lib.xmpp.chatState.js
```diff
@@ -130,9 +130,10 @@ jsxc.xmpp.chatState.onPaused = function(ev, jid) {
...
+      var duration = parseFloat(durationValue) || 0;
 
-      if (el.css('transition-duration').match(/s$/)) {
+      if (durationValue.match(/s$/)) {
```
ChristianTacke: I don't know, what values you expect anyway, but as I understand the above mentioned spec, there's also the option of "ms". So maybe: `.match(/[^m]s$/)`?

template/settings.html
ChristianTacke: Should I file an issue requesting that?

sualko: Yes, that would be nice.

src/jsxc.lib.xmpp.chatState.js
sualko: I think it doesn't matter. `0s` would be maybe cleaner, but the result is the same :smile:

src/jsxc.lib.xmpp.chatState.js
sualko: Great catch. Will fix it asap.

sualko:
Thanks to everyone you reviewed or contributed to this pr. It is really fun to work with you guys.

Gruntfile.js
sualko: f33dd0945481ed259901b9c802e97819dfe6c3ed

src/jsxc.lib.xmpp.chatState.js
mmoqui: done. I just submitted an issue about the translation problem

Gruntfile.js
ChristianTacke: That looks already a lot better.

template/settings.html
ChristianTacke: done: #434

ChristianTacke:
Thanks! Yes it's cool to work with all of you!

Issue #341, opened 2016-06-16
module HumHub


Andrea11:
Hi,
Given that I am a beginner with jsxc, I would like to develop a module for [HumHub](https://github.com/humhub/humhub). However I have a question about jsxc :

Can I customise the GUI (Hidden settings panel, remove the logout action, remov  audio and video calls, etc.)? If so, where should I look at?

Thanks

sualko:
Currently this feature is not supported, but it is on my todo list.

kousu:
Does jsxc support MUC Bookmarks? I see the right words in the backend code, but it doesn't seem to be respecting my bookmarks. I'm using oJSXC + prosody 10.0.0 + Nextcloud 13.0.0 Beta 1.

Gajim joins my room when I sign in, oJSXC doesn't.

kousu:
Oh nevermind #645 shows that @sualko is working on this right now. It's a bug in the UI; but the plumbing it all working.

PR #432, opened 2017-01-09
[WIP] Message Archive Management (XEP-0313)

### LINKED ISSUE #150: See complete history (XEP-0313)
I really like your app, but there is one thing missing.
I already compiled and installed a new version of ejabberd, to get mod_carboncopy ([XEP-0280](http://xmpp.org/extensions/xep-0280.html)) and mod_mam ([XEP-0313](http://xmpp.org/extensions/xep-0313.html))

I also activated Carbon Copy in jsxc-owncloud Settings.
But I have two problems:
- I can only access the last ~10 messages and not the whole history
- The history of my laptop is only on my laptop and not on my desktop pc

And idea? :)
##

singh1114:
fix #150 
Ref: http://xmpp.org/extensions/xep-0313.html


src/jsxc.lib.gui.js
```diff
@@ -2360,6 +2360,7 @@ jsxc.gui.window = {
     */
    open: function(bid) {
       var win = jsxc.gui.window.init(bid);
+      jsxc.xmpp.mam.getHistory(bid);
```
sualko: Please use an event handler for this. I think the appropriate event would be `init.window.jsxc`.

src/jsxc.lib.xmpp.mam.js
```diff
@@ -27,14 +27,111 @@
...
+
+  //self.getHistory(userid, count, before);
+};
+
+
+jsxc.xmpp.mam.getHistory = function(bid) {
+  /*
+  To include the whole functionality of strophe-plugin we used self
+  */
+  var self = jsxc.xmpp.conn.mam;
```
sualko: `self` usually points to the home namespace. In this case it would be `jsxc.xmpp.mam`.

sualko: You have to make sure, that the connection has already be established. See e.g. [jsxc.lib.xmpp.chatState.js#L17](https://github.com/jsxc/jsxc/blob/master/src/jsxc.lib.xmpp.chatState.js#L17)

src/jsxc.lib.xmpp.mam.js
```diff
@@ -27,14 +27,111 @@
...
+        <no-store xmlns="urn:xmpp:hints"></no-store>
+      </message>
+      
+      */
+      var direction;
+
+      /*
+      Check the direction of the message i.e. In which direction messages are exchanged.
+      */
+      if(bid === $(message).find("forwarded message").attr("to")){
```
sualko: Please use also the xml namespace to identify the forwarded element. See [jsxc.lib.xmpp.js#L789](https://github.com/jsxc/jsxc/blob/master/src/jsxc.lib.xmpp.js#L789).

src/jsxc.lib.xmpp.mam.js
```diff
@@ -27,14 +27,111 @@
...
+        "type": "plain",
+        "bid": bid,
+        "direction": direction,
+        "msg": msgBody
+      };
+
+      //console.log(message);
+      //console.log(mamUid);
+
+      jsxc.storage.setUserItem('msg', mamUid, msg);
```
sualko: This shouldn't be needed.

src/jsxc.lib.xmpp.mam.js
singh1114: Should I use 'this' than?

src/jsxc.lib.xmpp.mam.js
singh1114: OK. This will be removed.

src/jsxc.lib.xmpp.mam.js
singh1114: This is always equal to "<forwarded xmlns="urn:xmpp:forward:0">". What do we need to check here?

src/jsxc.lib.xmpp.mam.js
sualko: ???

src/jsxc.lib.xmpp.mam.js
sualko: In the future there could be an element with the same name, but different behavior.

singh1114:
I can't get anything on the sorting of the messages. Please help!

sualko:
If there is no sorting you have to implement it.

singh1114:
I have a doubt up here. I know the postMessage function can handle a single message at a time. So, how can I sort a single message?( Please tell me if I am wrong) 

Also, used a logic like this....

I tweaked the code of getHistory function such that it can whenever a user clicked on the someone from the roster list, all the messages are stored in a list and at the end, they are transferred to the postMessage function.

I wasn't able to get the required output but I want to ask is it the right way of solving the problem?

Another way is to create a function for sorting and when all the messages are stored in a data structure by the function getHistory they can be passed to that sorting function and finally they can be passed one by one to the postMessage. 

Also if someone can use the code written by me and see if the messages of MUC were retrieved or not it would be great.

sualko:
I would tweak the `_postMessage` function to work in a similar fashion than `jsxc.gui.roster.insert`.

skyfox675:
Quick feature question: Are there plans to add an option either in settings (global) or on a per conversation/chat basis to set history (On | Off | Server Default | Recipient Setting)? This is a very nice feature, not sure the difficult required to implement, just thought I would ask.

Thanks!

sualko:
@skyfox675 Thanks for your request. We will keep this in mind.

singh1114:
@sualko Can you check the sortMessages function. Have I done it correctly?

src/jsxc.lib.gui.js
```diff
@@ -2644,20 +2644,57 @@ jsxc.gui.window = {
...
+	            direction: jsxc.Message.SYS,
+	            msg: '42'
+	         }));
          }
       }
+      return message;
+   },
+
+   sortMessages: function(message){
+      var messageArray = jsxc.options.allMessages;
```
sualko: Why do you save messages in the options object?

src/jsxc.lib.gui.js
```diff
@@ -2644,20 +2644,57 @@ jsxc.gui.window = {
...
+      var length = messageArray.length;
+      var position = length;
+      var firstMessage;
 
+      do{
+         if (typeof messageArray[position-1] === 'undefined') {
+            firstMessage = true;
+         }
+         if (isNaN(message.stamp)) {
+            length = messageArray.length;
```
sualko: ???

src/jsxc.lib.options.js
```diff
@@ -5,6 +5,12 @@
  */
 jsxc.options = {
 
+   /**
+    * Declare empty list to be used for sorting the messages 
+    *  
+   */
+   allMessages: [],
```
sualko: The options object is only for options and not arbitrary data.

sualko:
> Have I done it correctly?

I don't think so. It would be probably more elegant if you would handle the sorting at [jsxc.lib.gui.js#L2803](https://github.com/jsxc/jsxc/blob/master/src/jsxc.lib.gui.js#L2803).

We also have to think about delayed messages, because with the new sorting function these will be maybe not visible to the user.

src/jsxc.lib.gui.js
singh1114: I was trying to do it in some other ways. For that, I initialized an empty data structure in a function. So whenever the function ran it again got initialized to empty.

For example: say I am declaring a variable in sortMessage function.
`sortMessages: function(message) {
    var messageArray = []
}`
So whenever the function ran it again got initialized to empty deleting all the messages. 

If you have any other idea please share.

src/jsxc.lib.gui.js
singh1114: So when I working with the messages I found that the last incoming message is not a real message it is just a system message which says that all messages have been encountered. I also found that the stamp of the last message returns NaN. So I am using that to check few things.

src/jsxc.lib.options.js
singh1114: Please suggest something.

src/jsxc.lib.gui.js
sualko: There has to be another solution, because this is :-1:.

src/jsxc.lib.gui.js
sualko: I mean, length was defined a couple of lines above and I think it is never changed.

sualko:
Maybe you should start with some simpler tasks, like our new [starter issues](https://github.com/jsxc/jsxc/labels/starter), to get familiar with the code.

singh1114:
How about #341...

sualko:
I think we close this pr. Thanks for your effort.

PR #493, opened 2017-03-17
Message Archive Management (XEP-0313)

### LINKED ISSUE #150: See complete history (XEP-0313)

sualko:
fix #150 
ref: https://xmpp.org/extensions/xep-0313.html

TODO:
- [x] namespace change?
- [x] [XEP-0334: Message Processing Hints](https://xmpp.org/extensions/xep-0334.html) for encrypted and maybe chat state messages.
- [x] slave tabs
- [x] testing


skyfox675:
@sualko Question about your implementation of XEP-0334, Section 4.2 outlines _no-store_ tags to indicate to the MAM component to not store messages. Was this part of your integration?

If not: Would it be possible to feature request for a option to be added to enable/disable the _no-store_ tag in the "div.jsxc_more" in chat menu?

sualko:
@skyfox675 `no-store` hints are optional and I use them currently only for OTR messages, but we always appreciate pull requests :wink:.

PR #502, opened 2017-03-24
A little bug: user name missing in the composing message

mmoqui:
Fix the following bug: when one user is writing a message, its name is missing in the composing message in the chat/groupchat window of the other(s) user(s)


sualko:
We should find a way to add the name only in rooms, because in a 2 person conversation this is not needed.

mmoqui:
I updated the code so that the name of the user(s) composing a message is(are) displayed only for groupchat. I rebased my changes to have only one commit.

sualko:
> I rebased my changes to have only one commit.

No need, because this can easily be done via githubs web ui. Thanks for your effort.

PR #505, opened 2017-04-06
Non translated text about the receipt of a message from an unknown sender

mmoqui:
fix the i18n key used when informing about a message coming from an unknown sender


sualko:
Thanks

Issue #470, opened 2017-02-27
Message to new contact


komantsev:
Hi! I use jsxc client with my OpenFire server.
I have 2 users (ivan001, ivan002).
Both users is online (**is important**).

Ivan002 (I2) adds ivan001 (I1) to contact list and sends him a message.
I1 received 2 notifications: add contact request and message from unknown user.
I1 clicks first notification and adds I2 to contact list.
I1 clicks second notification, accept request but he don't view a message from I2.
Messages from I1 to I2 is successfully delivered, but from I2 to I1 - not.

Page refresh (F5) fix this bug.

Here is video: https://www.youtube.com/watch?v=IuC3l1mIzZA

sualko:
Thanks for reporting. Can you check the javascript console from your right browser window?

komantsev:
Right window:
> State changed to INITIATING
> jsxc.js:185 State changed to PREVCONFOUND
> jsxc.js:185 check master
> jsxc.js:185 I am master.
> jsxc.js:185 State changed to ESTABLISHING
> jsxc.js:185 Try to attach
> jsxc.js:185 SID: 45l4wz5qvz
> jsxc.js:185 ATTACHED: null
> jsxc.js:185 UI State changed to INITIATING
> jsxc.js:185 Request server capabilities
> jsxc.js:185 UI State changed to READY
> jsxc.js:185 DSA key loaded
> jsxc.js:185 State changed to READY
> jsxc.js:185 Waiting for server capabilities
> jsxc.js:185 [ERROR] [XMPP] 404 remote-server-not-found
> jsxc.js:185 close dialog
> jsxc.js:185 Update file transfer icons for ivan001@pushtip.com
> jsxc.js:185 webrtc.initWindow
> jsxc.js:185 Update file transfer icons for ivan001@pushtip.com
> jsxc.js:185 Presence (ivan001@pushtip.com): online
> jsxc.js:185 Update file transfer icons for ivan001@pushtip.com
> jsxc.js:185 Presence (ivan001@pushtip.com/example): online
> jsxc.js:185 Update file transfer icons for ivan001@pushtip.com
> jsxc.js:185 Update file transfer icons for ivan001@pushtip.com
> jsxc.js:185 Auto approve contact request, because he is already in our contact list.
> jsxc.js:185 close dialog
> jsxc.js:185 Update file transfer icons for ivan001@pushtip.com
> 

Left window:

> jsxc.min.js:13 State changed to INITIATING
[...]

PR #511, opened 2017-04-12
Add the missing saveMessage function in jsxc.storage.

mmoqui:
When a message from an unknown sender is received, jsxc expects this function to save the incoming message for further treatment. But this function is missing causing an error with incoming messages from unknown user.

This missing function was first implemented in the Silverpeas chat client built upon JSXC to test it. Now I put it into JSXC.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -9,7 +9,7 @@ module.exports = function(grunt) {
 
    // Project configuration.
    grunt.initConfig({
-      github: grunt.file.readJSON('.github.json'),
+      //github: grunt.file.readJSON('.github.json'),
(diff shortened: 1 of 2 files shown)


sualko:
Where is this function be called? It was removed couple of versions before and I don't find it in the current code.

mmoqui:
Oups. I didn't remark the call in jsxc.lib.xmpp.js was removed recently (March 7, fix #470) and then the implementation of the missing function was not more necessary.
Sorry. I close then my PR.

sualko:
No problem. Thanks for your effort anyway. Keep up the good work!

PR #513, opened 2017-04-18
Take into account the 'RFC 6121 / No Available or Connected Resource' with an offline storage disabled

mmoqui:
In some circumstances, it is 'normal' to receive errors from a message sending. So, it is fine to not disturb the user with the reporting of such errors.
In this PR, we take into account the error is of type 'service unavailable'. Such an error is sent back to the message sender when the offline storage is disabled either by configuration or by the kind of the message (a message carrying a chatstate for example) and the recipient isn't connected (hence the targeted resource isn't available/connected). If the message that was sent carries a chatstate data, then the error isn't reported to the user (the sender). Otherwise, for other kinds of messages, he's informed its message won't be received by the recipient (as the offline storage is disabled). With this fix, the user won't be reported about message sending failures while he composes its message.

According to the XEP 0085, messages carrying chatstate data shouldn't be stored offline and according to the RFC 6121, when a recipient isn't connected, the message should be stored offline or an error has to be sent back. In the case of the chatstate notification,  as they shouldn't be stored offline, the 'normal' behavior for an XMPP server that want to satisfy the RFC is then to send back an error response. With this fix, I try to take into account this behavior that was first observed with ejabberd.


mmoqui:
I created also a PR in the Strophejs plugin chatstates to take into account such error messages coming from a chatstate notification (https://github.com/strophe/strophejs-plugin-chatstates/pull/1)

sualko:
https://tools.ietf.org/html/rfc6121#section-8.5.2.2

> If there are no available resources or connected resources associated
   with the user...

How should this happen in a group chat? If you send a message to a room, at least you are online/reachable.

src/jsxc.lib.muc.js
```diff
@@ -1421,6 +1421,14 @@ jsxc.muc = {
             direction: jsxc.Message.SYS,
             msg: $.t('message_not_send_not-acceptable')
          });
+      } else if ($(message).find('service-unavailable').length > 0) {
+         if ($(message).find("[xmlns='http://jabber.org/protocol/chatstates']").length === 0) {
+            jsxc.gui.window.postMessage({
+               bid: room,
+               direction: jsxc.Message.SYS,
+               msg: $.t('message_not_send_resource-unavailable')
```
sualko: The error message is confusing, because in a muc resources are used as nicknames.

mmoqui: The problem comes from only with peer-to-peer chat, not with muc. This error response should never comes for a muc.

mmoqui:
> How should this happen in a group chat? If you send a message to a room, at least you are online/reachable.
Yes. The problem occurs only for peer-to-peer chat. Sorry to not being more explicit about the context of the problem.

PR #575, opened 2017-07-04
add code of conduct

sualko:
@MarcelWaldvogel @LEDfan @DanScharon your opinions on this topic?


DanScharon:
I think it's a good idea. I personally would drop 'race' from the list as there already is ethnicity listed, but it's no showstopper for me.

LEDfan:
This is OK for me :+1:, especially since it's a wide used one!

Issue #591, opened 2017-08-01
entity discovery strophe bug (discovery requests not preserving id)


SproulDog:
### Expected behavior
Entity discovery strophe bug (discovery requests not preserving id). Expected behaviour is that presence/caps received by client will result in discovery query for identity/features of that node/hash (unless otherwise recieved by the client).

### Actual behavior
Missing return statement results in ophaned callback when JSXC attempts discovery of other entities,
The observed misbehaviour is that it does not seem that the discovery process is issued/succeeds and the identity/feature-set is not correct for the client/stream entity. ie: if you examine the "more info" for the contact, the identity portion of the contact details is blank (and when debugging, it seems that the featureset for the stream/client are also empty (despite the node and version caps hash being cached).

### Steps to reproduce the behavior
Strophe Plugin dependency (disco) contains an oversight:
`info: function(jid, node, success, error, timeout)`
needs to return id from `this._connection.sendIQ(info, success, error, timeout)`

`_requestCapabilities: function(to, node, ver)` calls it expecting to receive the id (see below):
```
var id = this._connection.disco.info(to, node + '#' + ver);
```

but the id is not returned.

### Suggested Change
Change:
`this._connection.sendIQ(info, success, error, timeout)`
To:
`return this._connection.sendIQ(info, success, error, timeout)`

Modified in: https://github.com/jsxc/jsxc/pull/592

As a result, the JSXC requestor does not receive the correct result from the issued discovery request.
### Environment
- **JSXC version: jsxc v3.2.1 - 2017-06-01
- **Host system and version: N/A
- **Browser vendor and version: N/A 
- **Any browser plugins enabled? N/A
- **XMPP server vendor and version: N/A 
- **Is your XMPP server working with other clients as expected? N/A

PR #592, opened 2017-08-01
Update strophe.disco.js

SproulDog:
The id from sendIQ should be returned from the info request to pair with the response.
Reference: https://github.com/jsxc/jsxc/issues/591


sualko:
Thanks for your contribution. Great catch :+1:

PR #662, opened 2018-01-22
fix unread flag for mam

usanjar:


sualko:
Can you give me a short reminder what this pr should fix?

PR #737, opened 2018-10-10
Prevent undefined 'last' key in array loop when marking quotations

jlemangarin:
When marking quotations, we loop into the typed lines.

In some configurations, the for loop set the lineNumber to 'last' at the end, so the line can't be retrieved and an undefined error is triggered in the "line.indexOf' part.

### DIFF
--- a/src/Message.ts
+++ b/src/Message.ts
@@ -269,6 +269,11 @@ export default class Message implements Identifiable, IMessage {
       let inQuote = false;
 
       for (let lineNumber in lines) {
+         if (!lines.hasOwnProperty(lineNumber)) {
+            //This value comes from Array prototype
+            continue;
+         }
+
          let line = lines[lineNumber];
 
          if (line.indexOf('&gt;') === 0) {


sualko:
Can you give an example of such a conversation?

Travis is probably reporting a style error (we use single quotes) and a commit message error. It would be awesome if you could satisfy Travis, otherwise I will do it. No problem.

Thanks for your pr.

jlemangarin:
Yep, there is an exemple (really simple, I just have to type a message :) )

capture d ecran de 2018-10-10 17-09-55
Try to send it
capture d ecran de 2018-10-10 17-10-02
Same with the fix
capture d ecran de 2018-10-10 17-11-04
Because
capture d ecran de 2018-10-10 17-17-49

Sorry but I can't satisfy Travis even if I replace the double quotes with simple quotes :'(

jlemangarin:
Made it using yarn autofix script. Travis build passed ! :)

sualko:
After reading [why for in is a bad idea for arrays](https://stackoverflow.com/questions/500504/why-is-using-for-in-with-array-iteration-a-bad-idea) , I think the proper solution is to use an "old school" `for` loop. Can you do that? It would also be awesome if you could [squash your commits](https://github.com/todotxt/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit) and change your commit message according to https://www.conventionalcommits.org/.

jlemangarin:
Nope
capture d ecran de 2018-10-12 09-36-12

I'm in the latest stable release (69.0.X). Maybe a consequence of : 
https://www.chromestatus.com/feature/6629507075145728 (https://www.chromestatus.com/features/schedule)

jlemangarin:
Just seen your last comment. Ok I'll try !

sualko:
> Nope

I guess `[].last` will print some function, instead of a native prototype function like `[].splice`.

> Ok I'll try !

Awesome :tada:

jlemangarin:
Aw, yes you're right. Anyway it should be better to not iterate in array like that. :)

sualko:
> Commit squashed and conventionalized !

Thanks a lot! :fireworks: :champagne:

PR #742, opened 2018-11-09
Fix the rendering of messages of the current user when a room window is just opened

mmoqui:
onGroupChatMessage is invoked when a room window is opened by jsxc. Each message passed as argument to this function seams to come from
the XMPP service (ejabberd in my case) and among them there is the
messages from the current user. The change in the function is to take
care of the messages coming from the current user: in this case, the
direction is explicitly set to jsxc.Message.OUT and the message is
then simply rendered.

Replace some explicit value 'in' or 'out' for message direction by
their respective constant jsxc.Message.IN and jsxc.Message.OUT.
Rename the function jsxc.gui.window._postMessage by
jsxc.gui.window.renderMessage.

### DIFF
--- a/src/jsxc.lib.gui.js
+++ b/src/jsxc.lib.gui.js
@@ -2659,7 +2659,7 @@ jsxc.gui.window = {
          });
       }
 
-      if (message.direction === 'in' && !jsxc.gui.window.get(message.bid).find('.jsxc_textinput').is(":focus")) {
+      if (message.direction === jsxc.Message.IN && !jsxc.gui.window.get(message.bid).find('.jsxc_textinput').is(":focus")) {
          jsxc.gui.unreadMsg(message.bid);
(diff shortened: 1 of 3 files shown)


mmoqui:
Any comment about my PR?

sualko:
I was currently working on your pr :laughing:. As you can see in cbb300eb3e49c8235fca8b73a24dea0cb66c5f9e and a369479a347b08215e889bb4456019e3f3a7b217 I splitted your pull request into multiple commits. Currently I'm looking at the original cause of your action and as I said in the other issue, we can't assume that those mam message are from you. Therefore I can't accept those changes, but I will add a fix which should solve your issue. Just wait some minutes :wink:

sualko:
I close this in favor of d62604a. Thanks for your contribution :+1:

PR #752, opened 2018-12-07
User Nickname (XEP-0172)

### LINKED ISSUE #504: Support for XEP-0172: User Nickname
This would be a very nice feature, to have the name from vcard-temp show up automatically.

Fr3yn:
ref: https://xmpp.org/extensions/xep-0172.html
fix #504

TODO:

- [x] Read into XEPs and code
- [x] Implement Nickname Change with vCard
- [x] Include nickname into presence subscription request
- [x] Publish changed nickname to PEP Node
- [x] Show nickname in Roster


sualko:
The commit message of 42f04dada91c7855fe189bd9bdc0917d5085b8ac doesn't match the change :wink:. Was this intended?

Fr3yn:
> The commit message of [42f04da](https://github.com/jsxc/jsxc/commit/42f04dada91c7855fe189bd9bdc0917d5085b8ac) doesn't match the change 😉. Was this intended?

Oh, no this was indeed not intended! Seems that the changes in the Contact.ts were not committed... :)

Fr3yn:
We ran into an issue while working on this feature: How do we get the user's nickname, meaning where do they type it in? We were thinking maybe a additional menu item would work that would open a dialog where the user could change his or her nickname. We wanted to ask for second opinions and ask another question: where are the items of the jsxc menu defined? There's the menu.hbs but I don't really understand how it works. We really appreciate any help, thanks in advance :).

MarcelWaldvogel:
My wild guesses without knowing the 4.x innards:

Nickname entry: In the same dialog where the avatar picture is entered (if the 4.x branch already has that feature).

`menu.hbs`: Is it referenced anywhere? What file/code/data is created from it?

Fr3yn:
> My wild guesses without knowing the 4.x innards: Nickname entry: In the same dialog where the avatar picture is entered (if the 4.x branch already has that feature). `menu.hbs`: Is it referenced anywhere? What file/code/data is created from it?

Thanks for the quick reply! I can't seem to find the dialog where the avatar picture is set, neither in 3.x nor in 4.x, where do I find it? The menu.hbs seems to be the template that creates the menu but its only a few lines long and I can not find the file where the entries are set.

Fr3yn:
Never mind. I think I found what I needed. Thanks and happy new year :).

Fr3yn:
Hello everybody,

so we are implementing the PEP support for nicknames right now and seem to be stuck at the following point: we publish the PEP event when updating the nickname right now, like specified in XEP-0172. This seems to work, the debugging mode gives us the following output: 

<body rid="369778486" xmlns="http://jabber.org/protocol/httpbind" sid="364f5e78-5033-4631-93b9-c123c3fbcb5e">
<iq type="set" xmlns="jabber:client" id="fc596906-ab28-444f-8eec-143bff7d65a6:sendIQ">
<pubsub xmlns="http://jabber.org/protocol/pubsub">
<publish node="http://jabber.org/protocol/nick"><item>
<nick xmlns="http://jabber.org/protocol/nick">
Nickname
</nick>
</item>
</publish>
</pubsub>
</iq>

and it seems to generate this afterwards: 
<iq xmlns="jabber:client" to="my-mail@uni-konstanz.de/p86X4_nk" id="fc596906-ab28-444f-8eec-143bff7d65a6:sendIQ" type="result">
<pubsub xmlns="http://jabber.org/protocol/pubsub">
<publish node="http://jabber.org/protocol/nick">
<item id="05acf36b-cf7f-44c0-a937-a62f40ab252c"/>
</publish>
</pubsub>
</iq>

Our contacts do not receive anything. According to the XEP 0172, we and the contacts who subscribed to us, should receive something like this:

<message from='narrator@moby-dick.lit'
         to='starbuck@moby-dick.lit'
         type='headline'
         id='foo'>
  <event xmlns='http://jabber.org/protocol/pubsub#event'>
    <items node='http://jabber.org/protocol/nick'>
      <item>
        <nick xmlns='http://jabber.org/protocol/nick'>CallMeIshmael</nick>
      </item>
    </items>
  </event>
  <addresses xmlns='http://jabber.org/protocol/address'>
    <address type='replyto' jid='narrator@moby-dick.lit/pda'/>
  </addresses>
</message>

But we do not. Is this expected behavior?

Fr3yn:
> I guess you have to subscribe to the node. Just look at the pep service class.

Okay, but what do we put as the handler?

sualko:
> Okay, but what do we put as the handler?

The function/method which should process the event.

Fr3yn:
Okay, it seems like we are stuck at some point. We subscribe to a contact when adding him to our roster, we send PEP updates when changing the nickname, but somehow, it does not seem to be transmitted as we don't receive any updates. Any help would be greatly appreciated :).

Fr3yn:
We might have found something that could cause the problem described above: when publishing the nickname to PEP, the 'to' and 'from' value of the generated notification is the same (the jid of the sender) while in the XEP it is specified that it should generate notifications for every contact that has subscribed to us. Is that behaviour expected?

sualko:
What is the current state of your pr?

Fr3yn:
We are reviewing the code and will ask for a merge as soon as that's done.On 15 Feb 2019 13:42, Klaus <notifications@github.com> wrote:What is the current state of your pr?

Fr3yn:
We have reviewed the code and would like to ask you for a merge :).

src/connection/xmpp/Connection.ts
```diff
@@ -7,20 +7,29 @@ import Account from '../../Account'
 import { AbstractConnection, Presence, STANZA_IQ_KEY, STANZA_KEY } from '../AbstractConnection'
 import XMPPJingleHandler from './JingleHandler'
 import Client from '../../Client'
+import Contact from '@src/Contact';
 
 export default class XMPPConnection extends AbstractConnection implements IConnection {
    private handler: XMPPHandler;
-
```
sualko: Why did you remove this blank line?

src/connection/xmpp/Connection.ts
```diff
@@ -7,20 +7,29 @@ import Account from '../../Account'
...
       this.handler = new XMPPHandler(account, connection);
       this.handler.registerHandler();
       NS.register('METADATA_NOTIFY', 'urn:xmpp:avatar:metadata+notify');
 
+      this.getPEPService().subscribe('http://jabber.org/protocol/nick', nickHandler);
+
       this.processPendingStorageConnections();
       this.listenToStorageConnections();
 
+      function nickHandler(stanza): boolean {
```
sualko: Inline function functions should be avoided.

src/connection/xmpp/Connection.ts
```diff
@@ -7,20 +7,29 @@ import Account from '../../Account'
...
+         let nickname: string = $(stanza).find('nick').text();
```
sualko: Is there no more specific query string? No namespace?

src/connection/xmpp/Connection.ts
```diff
@@ -7,20 +7,29 @@ import Account from '../../Account'
...
+         let peerJID = new JID($(stanza).attr('from'));
```
sualko: Now stanza is parsed the second time.

src/connection/xmpp/handlers/chatMessage.ts
```diff
@@ -22,13 +22,19 @@ export default class extends AbstractHandler {
 
       let peerJid = new JID(messageElement.getOriginalFrom());
       let peerContact = this.account.getContact(peerJid);
+      let nickname: string = $(stanza).find('nick').text();
+      let oldName: string = peerContact.getNickname();
+      if ((nickname !== undefined) && (oldName !== nickname)) {
```
sualko: Why is it not allowed to remove a nickname?

src/connection/xmpp/handlers/headlineMessage.ts
```diff
@@ -8,6 +8,12 @@ export default class extends AbstractHandler {
       let fromJid = new JID(stanza.getAttribute('from'));
       let connection = this.account.getConnection();
       let myJid = connection.getJID();
+      let nickname: string = $(stanza).find('nick').text();
```
sualko: Same code as in chatMessage. Stay DRY.

src/ui/dialogs/contact.ts
```diff
@@ -5,14 +5,15 @@ import Client from '../../Client'
...
-
+let nick: string;
 let dialog: Dialog;
 
-export default function(username?: string) {
+export default function(username?: string, nickname?: string) {
    username = (typeof username === 'string') ? username : undefined;
-
+   nickname = (typeof nickname === 'string') ? nickname : undefined;
+   nick = nickname;
```
sualko: ???

src/ui/dialogs/settings.ts
```diff
@@ -126,17 +127,86 @@ class ConnectionSection extends Section {
...
+
+      let errorElement = $(`<div class="jsxc-alert jsxc-alert--warning jsxc-hidden"></div>`);
+
+      contentElement.append(explanationElement);
+      contentElement.append(nicknameElement);
+      contentElement.append(submitElement);
+      contentElement.append(errorElement);
+
+      contentElement.find('input').on('input', function() {
+         let nickname = nicknameElement.find('input').val();
```
sualko: unused

src/ui/dialogs/settings.ts
```diff
@@ -126,17 +127,86 @@ class ConnectionSection extends Section {
...
+         this.account.getConnection().changeNickname(nickname).then(() => {
+            Log.debug('Nickname was changed');
+         }).catch((errStanza) => {
+
+            if (errStanza === TypeError) {
+               throw errStanza;
+            }
+
+            errorElement.removeClass('jsxc-hidden');
+            errorElement.text('Server error. Nickname was not changed.');
```
sualko: Do you reset this message somewhere?

src/connection/xmpp/handlers/presence.ts
```diff
@@ -98,7 +101,7 @@ export default class extends AbstractHandler {
          description: 'from ' + jid.bare,
          type: NOTICETYPE.contact,
          fnName: NOTICEFUNCTION.contactRequest,
-         fnParams: [jid.bare]
+         fnParams: [jid.bare, nickname]
```
sualko: Is this used somewhere?

src/Account.ts
```diff
@@ -66,6 +66,7 @@ export default class Account {
 
          let oldSessionId = this.getStorage().getItem('sessionId');
          this.getStorage().setItem('sessionId', this.sessionId);
+         this.getStorage().generateKey('nickname');
```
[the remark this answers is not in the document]
Fr3yn: We store our own nickname in the local storage and have to generate a key for that.

src/connection/xmpp/handlers/presence.ts
Fr3yn: It is used to keep the nickname we received from the presence request in the cache,  we later use it in the dialogs/contact.ts file.

src/Account.ts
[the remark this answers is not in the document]
sualko: You should really look at this function. It's not doing what you are thinking.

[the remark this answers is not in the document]
Fr3yn: You're right, I sould've looked at the implementation first. In that case it is unnecessary.

src/connection/AbstractConnection.ts
```diff
@@ -227,6 +235,31 @@ abstract class AbstractConnection {
       return this.sendIQ(iq);
    }
 
+   public changeNickname(newNickname: string): Promise<Element> {
```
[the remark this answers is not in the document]
Fr3yn: AbstractConnection.changeNickname() is meant to handle the change of the nickname in the vCard (as a connection is required for this to change). Account.setNickname() changes the locally stored nickname, both are called when clicking the submitting button in the nickname settings page.

src/connection/xmpp/Connection.ts
Fr3yn: We tried defining the function outside the constructor, but this results in a TypeError because 'account' allegedly is undefined (although it still works). Is this solution preferred although it results in errors?

sualko: Just fix the error :wink:. Test which value `this` has and use maybe an arrow function.

Fr3yn: Like this?

src/connection/xmpp/handlers/headlineMessage.ts
Fr3yn: How would I reuse code from the method of another class?

sualko: Put the code in it's own class for example. Therefore all nickname related code would be at one place. I think it would be a good semantic to group the code, because every contact has a nickname and it can be read and updated. Even every account has a own contact.

Fr3yn:
Okay, should we put the class into the plugins folder, or simply src?On 21 Feb 2019 09:48, Klaus <notifications@github.com> wrote:@sualko commented on this pull request.



In src/connection/xmpp/handlers/headlineMessage.ts:
> @@ -8,6 +8,12 @@ export default class extends AbstractHandler {
       let fromJid = new JID(stanza.getAttribute('from'));
       let connection = this.account.getConnection();
       let myJid = connection.getJID();
+      let nickname: string = $(stanza).find('nick').text();

Put the code in it's own class for example. Therefore all nickname related code would be at one place. I think it would be a good semantic to group the code, because every contact has a nickname and it can be read and updated. Even every account has a own contact.

sualko:
I think it's not possible to implement this XEP as plugin, therefore just put it into the src folder.

Fr3yn:
I think some of the functionality can not be put into the nickname class, so we would like to ask you to merge our changes please :).

sualko:
> I think some of the functionality can not be put into the nickname class

But you can resolve all comments and remove dead code from your pull request. In this state it's unmergeable.

src/connection/xmpp/handlers/headlineMessage.ts
```diff
@@ -8,6 +8,13 @@ export default class extends AbstractHandler {
       let fromJid = new JID(stanza.getAttribute('from'));
       let connection = this.account.getConnection();
       let myJid = connection.getJID();
+      let nicknameObject = this.account.getContact(myJid).getNicknameObject();
+      let nickname: string = nicknameObject.getNickFromStanza(stanza);
+      let oldName: string = nicknameObject.getNickname();
+
+      if ((nickname !== undefined) && (oldName !== nickname)) {
+         this.account.getContact(fromJid).getNicknameObject().setContactNickname(nickname);
+      }
```
sualko: Stay DRY. See `chatMessage.ts`.

src/plugins/bookmarks/BookmarkProvider.ts
```diff
@@ -134,7 +134,7 @@ export default class BookmarkProvider extends ContactProvider {
 
    private initBookmarkContact(bookmark: RoomBookmark, service: AbstractService): IContact {
       let contact = this.createContact(bookmark.getJid());
-      contact.setNickname(bookmark.getNickname());
+      contact.getNicknameObject().setContactNickname(bookmark.getNickname());
```
sualko: The bookmark is storing my own nickname inside the room. So I think we have to distinguish here.

src/ui/RosterItem.ts
```diff
@@ -174,6 +174,7 @@ export default class RosterItem {
       let inputElement = this.element.find('input');
 
       this.contact.setName(<string> inputElement.val());
+      this.contact.getNicknameObject().setNickVisible(false);
```
sualko: I guess this should enforce the user defined name. But this would be easier with priorities. E.g. `getName() || getNickname() || getBareJid()`.

src/ui/dialogs/settings.ts
```diff
@@ -7,6 +7,7 @@ import List from '../DialogList'
 import ListItem from '../DialogListItem'
 import AvatarSet from '../AvatarSet'
 import Log from '../../util/Log'
+//import Contact from '@src/Contact';
```
sualko: ???

src/connection/xmpp/Connection.ts
```diff
@@ -7,23 +7,44 @@ import Account from '../../Account'
 import { AbstractConnection, Presence, STANZA_IQ_KEY, STANZA_KEY } from '../AbstractConnection'
 import XMPPJingleHandler from './JingleHandler'
 import Client from '../../Client'
+import Nickname from '@src/Nickname';
+
+let acc;
```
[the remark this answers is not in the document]
Fr3yn: What exactly?

src/connection/services/Roster.ts
```diff
@@ -78,10 +77,15 @@ export default class Roster extends AbstractService {
    }
 
    private sendSubscriptionRequest(jid: IJID) {
+      let nicknameObject = this.account.getContact().getNicknameObject();
+      let nickname = nicknameObject.getNickname();
+      let reference = nicknameObject.getNickRef();
```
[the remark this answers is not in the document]
Fr3yn: The nickname namespace does not seem to be included in the Strophe.NS method.

src/connection/xmpp/Connection.ts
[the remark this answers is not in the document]
sualko: A global variable with fuzzy name

src/connection/services/Roster.ts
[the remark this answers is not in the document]
sualko: `connection/xmpp/namespace.register`

src/connection/xmpp/handlers/headlineMessage.ts
Fr3yn: How would we reuse the code from chatMessage for this if statement?

PR #778, opened 2019-05-03
change WEB UI State to READY

pussbb:


sualko:
Thanks for the pull request, but I think we will remove the state machine until the final release. It's a holdover from the 3.x branch. Do you have any particular use case for it?

pussbb:
Hi,
Well not sure if its right way but in v3 I use event  ''ready.roster.jsxc" to change height of the JSXC application.
Well application has menubar and toolbar JSXC on show takes full height at the right side of the page but it hides some controls in main page. So when  ''ready.roster.jsxc" I change top position to lower and adding own div to toggle buddy list  (show/hide user but keep "jsxc_bottom jsxc_presence" element).
As next step I wanted to add vertical resizing for buddy list. 
Generally speaking with UI states (events) I know when elements is shown and can manipulate them to meet my UI requirements and features. 
Does not know for if JSXC v4 offers the same just started to work with it

sualko:
> I change top position to lower 

This can also be done via css, see https://github.com/nextcloud/jsxc.nextcloud/blob/dfb5454addd0a36181d686818c9803d14660c101/scss/_oc.scss#L19-L23

> adding own div to toggle buddy list

You could just check if `jsxc.numberOfCachedAccounts` is larger than zero after initializing jsxc, or add it after a successful connection attempt.

PR #802, opened 2019-08-07
fix: handle empty stream

pussbb:
When user starts video call in Jitsi Meet application at JSXC appears Incoming call notification.
User accepts it but if he wants to hang up(drop) call he wont be able to do it because argument stream is null and in console will be message which looks like 
``` can not get property getTracks from ...```

### DIFF
--- a/src/ui/VideoDialog.ts
+++ b/src/ui/VideoDialog.ts
@@ -259,6 +259,12 @@ export class VideoDialog {
    }
 
    private static stopStream(stream) {
+
+      if (!stream) {
+         Log.warn('Could not stop stream. Stream is null.');
+         return;
+      }
+
       if (typeof stream.getTracks === 'function') {
          let tracks = stream.getTracks();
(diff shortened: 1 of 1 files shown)


pussbb:
In general its not the only one error if someone will try room in Jitsi conference call but at least they do not harm for user interface (appears in console only)

sualko:
How are Jitsi Meet and JSXC interoperating? How can I reproduce this?

pussbb:
if apply pull request from me to add ability set config option for plugin you can use
https://gist.github.com/pussbb/e7b78f9b3f73d286d42c805f3cc5feb4
see details there. 
then please create room , allow add participants to all (from member to anyone) .

Without its hard, issue is that Jitsi do not allow uses their XMPP server. Working on Jitsi based application https://github.com/scalix/jitsi-meet but I have only production server I could give access but you would see other users. If you are ok I will try to setup demo server where not real people lives but it will took some time.

PR #804, opened 2019-08-08
PluginApi enhastments (plugin options)

pussbb:
Once again faced with limited Plugin functionality for developers. 
So this pull request allows Plugin to get settings via getOption method call and gives ability to set custom options for plugin.

 simple example usage
```javascript
let JSXC_MIN_VERSION = '4.0.0';
let JSXC_MAX_VERSION = '4.9999.0';

class GroupMeetingCall extends JSXC.AbstractPlugin {

   options;

   constructor(pluginAPI) {
      super(JSXC_MIN_VERSION, JSXC_MAX_VERSION, pluginAPI);

      this.options = {
         ...GroupMeetingCall.defaultOptions(),
         ...(pluginAPI.getOption(GroupMeetingCall.getName()) || {})
      };
      const invalidUrl = !this.getMeetUrl();
      pluginAPI.registerChatWindowInitializedHook((chatWindow) => {
         if (chatWindow.getContact().getType() !== 'groupchat' || invalidUrl) {
            return;
         }
         chatWindow.addMenuEntry('meet', 'Call', () => this.startCall(chatWindow));
      });
   }

   getMeetUrl() {
      return this.options.url;
   }

   static getName() {
      return 'GroupMeetingCall';
   }

   static defaultOptions() {
      return {
         'url': ''
      }
   }

   startCall(chatWindow) {
      let jid = chatWindow.getContact().getJid().node;
      let url = this.getMeetUrl();

      if (!url.endsWith('/')) {
         url += '/';
      }

      let a = document.createElement('a');
      a.href = url + jid;
      a.target = '_blank';
      a.rel = 'noopener noreferrer';
      a.click();
   }
}

// set options.
jsxc.addPlugin(GroupMeetingCall, {
    'url': 'https://meet.jitsi.si/'
});

// use default plugin options
// jsxc.addPlugin(GroupMeetingCall)
```

### DIFF
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -3,7 +3,7 @@ import Storage from './Storage'
 import { NoticeManager } from './NoticeManager'
 import PluginRepository from './plugin/PluginRepository'
 import Log from './util/Log'
-import Options from './Options'
+import Options, {IOptionData} from './Options'
 import PresenceController from './PresenceController'
 import PageVisibility from './PageVisibility'
 import ChatWindowList from './ui/ChatWindowList';
@@ -54,9 +54,12 @@ export default class Client {
       return '4.0.0';
    }
 
-   public static addPlugin(Plugin: IPlugin) {
+   public static addPlugin(Plugin: IPlugin, defaultOptions?: IOptionData) {
       try {
          PluginRepository.add(Plugin);
+         if (defaultOptions) {
+            Options.addDefaults(defaultOptions);
+         }
       } catch (err) {
          Log.warn('Error while adding Plugin: ' + err);
       }
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -58,7 +58,9 @@ export default class Client {
       try {
          PluginRepository.add(Plugin);
          if (defaultOptions) {
-            Options.addDefaults(defaultOptions);
+            let options = {};
+            options[Plugin.getName()] = defaultOptions;
+            Options.addDefaults(options);
          }
       } catch (err) {
          Log.warn('Error while adding Plugin: ' + err);
--- a/src/Options.ts
+++ b/src/Options.ts
@@ -8,7 +8,7 @@ import Client from './Client';
 
 const KEY = 'options';
 
-interface IOptionData {
+export interface IOptionData {
    [key: string]: any
 }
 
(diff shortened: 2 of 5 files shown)


pussbb:
Please note that its better to have ``options`` class member in AbstractPlugin.
```javascript
export abstract class AbstractPlugin {

   protected options;

   public static getName(): string {
      return null;
   }

   constructor(protected minVersion: string, protected maxVersion: string, protected pluginAPI: IPluginAPI) {
      if (!this.isSupportingClientVersion()) {
         throw new Error('This plugin doesn\'t support this client version');
      }

      this.options = pluginAPI.getOption((<any> this.constructor).getName());
   }
....
```

and plugin code becomes much more cleaner
```javascript
class GroupMeetingCall extends JSXC.AbstractPlugin {
    
   constructor(pluginAPI) {
      super(JSXC_MIN_VERSION, JSXC_MAX_VERSION, pluginAPI);
      console.log(this.options);
.... 
```
But there are a lot of plugins implement internally and adding such thing can break some functionality.

pussbb:
as an option following methods can be implemented in AbstractPlugin
```javascript
export abstract class AbstractPlugin {
   ...

   public getPluginOptions() {
      //TODO make sure that merge of object's is deep merge.
      return {
         ...(<any> this).getDefaultOptions(),
         ...(this.pluginAPI.getOption((<any> this).getName()) || {})
      };
   }

   protected getDefaultOptions() {
      return {};
   }
...
```
Does not sure what way is the best. So will leave for your consideration.

sualko:
Can you explain what you want to archive? Because I think the plugin doesn't need to have it's own options class, because it can just use the provided storage (`pluginAPI.getStorage()`).

sualko:
After looking at https://gist.github.com/pussbb/e7b78f9b3f73d286d42c805f3cc5feb4, I have an idea what you want. I like the idea and will probably modify your code a bit and merge it. Thanks a lot. It's a really cool plugin, which I would like to integrate into JSXC core if that's ok for you.

pussbb:
I dont mind if you integrate that plugin into JSXC.

sualko:
After looking at the plugin more closely I saw that this basically adds a link which opens a jitsi meet session. Thought it would be a real group call, but nonetheless, thanks for your contributions and the testing you did. It really helps to enhance JSXC :+1:

sualko:
After thinking about your pr, it's probably not that easy. Because we should handle the default, developer and user options similar to all other options in JSXC. It would also be nice to have an easy way to add new options to the settings menu. I think it's better to create a more powerful solution after the release of 4.0, than to merge it in the current state. In your case I would recommend to use a static variable to set the url. Something like the following:

```
GroupMeetingCall.url = '...';
jsxc.addPlugin(GroupMeetingCall);
```

I leave this pr open as a note to myself.

PR #807, opened 2019-08-19
feat: api to create and join muc rooms

jlemangarin:
feat https://github.com/jsxc/jsxc/issues/805

PR #809, opened 2019-08-21
feat: password sub element in bookmark storage for conference element

jlemangarin:
Ref : https://xmpp.org/extensions/attic/xep-0048-1.0.html part 2.2

The "password" sub-element should be stored when it's present in the contact data (refer to https://github.com/jsxc/jsxc/pull/807)

### DIFF
--- a/src/MultiUserContact.ts
+++ b/src/MultiUserContact.ts
@@ -179,6 +179,10 @@ export default class MultiUserContact extends Contact {
       return this.data.get('subject');
    }
 
+   public getPassword(): string {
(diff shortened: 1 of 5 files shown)


jlemangarin:
Missing storage of the password when retrieving the bookmark data so the password is erased from bookmark at reconnection. I'm on it

jlemangarin:
Commit completed, @sualko could you please review it ? :)

sualko:
Thanks for your contribution. I think we are now ready for the first release candidate.

PR #821, opened 2019-09-05
Remove underscore suffixes from i18n keys

ar-maged:


sualko:
Suffixes indicate that the translation is longer than the variable.

PR #826, opened 2019-09-10
fix:  use getId() for checking set encryption state

pussbb:
When trying to turn on / off some encryption plugin you will get in console 
```plain
Uncaught Error: Couldn't find omemo
    at PluginRepository.getEncryptionPlugin (PluginRepository.ts:81)
    at HTMLDivElement.eval (ChatWindow.ts:474)
    at HTMLDivElement.dispatch (jquery.min.js:2)
    at HTMLDivElement.y.handle (jquery.min.js:2)
```

OMEMO was not working - well messages was not encrypted and was send as plain text. 

since ```PluginRepository.getEncryptionPlugin ``` moved to method ```getID()``` so changed rest parts .

### DIFF
--- a/src/Contact.ts
+++ b/src/Contact.ts
@@ -269,7 +269,6 @@ export default class Contact implements IIdentifiable, IContact {
       if (state !== EncryptionState.Plaintext && !source) {
          throw new Error('No encryption source provided');
       }
-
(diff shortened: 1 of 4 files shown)


sualko:
Once more: Thanks :tada:

PR #827, opened 2019-09-10
fix: ignore components which does not support DiscoInfo on determining MUC services at XMPP server

pussbb:
Jitsi has speakerstats_component which is not support getting discoinfo and responses with 
```xml
<iq xmlns="jabber:client" type="error" to="as@site.com/jsxc-6e71fee0" from="speakerstats.site.com" id="4bcd4257-0c2a-423a-b458-60b3d9079b8d:sendIQ">
<error type="cancel">
    <service-unavailable xmlns="urn:ietf:params:xml:ns:xmpp-stanzas"/>
</error>
</iq>
```
which is stopping searching for MUC component domain.

Following change set allows to ignore ``service-unavailable`` for components which does not support disco items request and will re-throw ``stanza`` item in something else happend.

### DIFF
--- a/src/ui/dialogs/multiUserJoin.ts
+++ b/src/ui/dialogs/multiUserJoin.ts
@@ -116,14 +116,10 @@ class MultiUserJoinDialog {
             }).then((hasFeature) => {
                return hasFeature ? jid : undefined;
             }).catch((stanza) => {
-            const from = $(stanza).attr('from') || '';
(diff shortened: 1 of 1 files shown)


sualko:
According to [XEP-0045: Multi-User Chat](https://xmpp.org/extensions/xep-0045.html#entity) a MUC implementation MUST support Service Discovery (XEP-0030). Therefore we can safely ignore any failing disco request. Thanks for your pr.

PR #841, opened 2019-10-31
feat: adding support for RTL messages in chat

niryas:
- A simple addition to the textarea markup (`dir="auto"`) to allow the modern browser to automatically decide the direction of the text (LTR/RTL).

- Addition of same markup (`dir="auto"`) to the `<p>` tag of chat messages.

- Change of line-break code: instead of `<br />`, the lines are split with `<p>` tags so the direction markup can be applied per line. This allows the different lines of the same message to have a different direction (e.g., when writing one line of English and one of Hebrew/Arabic/etc).

MDN info about `dir="auto"`:
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir

### DIFF
--- a/src/Message.ts
+++ b/src/Message.ts
@@ -262,7 +262,7 @@ export default class Message implements IIdentifiable, IMessage {
       body = Utils.escapeHTML(body);
       body = await Message.formatText(body, this.getDirection(), this.getPeer());
 
-      return `<p>${body}</p>`;
+      return `<p dir="auto">${body}</p>`;
    }
 
    public getPlaintextEmoticonMessage(): string {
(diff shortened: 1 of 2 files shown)


sualko:
Thanks for your pr, but I think it needs some more work. I discovered the following issues:

- [ ] multi line quotes should be inside one p tag
- [ ] empty line (e.g. `foo\n\nbar`) is not visible

Can we assume that an empty line separates two paragraphs? Because in this case we could first replace two line breaks with a new paragraph and a single break would result in a normal line break.

sualko:
Is this the intended behavior? I would expect that both have the same direction.
2020-03-26 09 29 12 localhost 022cad846786

sualko:
Merged via babb556b8b91668e70c356dc156fe49dae2e2a6e. Thank you for your contribution. :tada:

PR #864, opened 2020-04-01
Remove sound files impossible to redistribute

linkmauve:
The ringtone sound file wasn’t licensed for redistribution as far as I
know, this prevents JSXC from being allowed to redistribute it without
written permission from the original author.

If you want to keep using this sound file you could try to contact the
author and see if they agree with you redistributing it, in the meantime
it’s safer both for the project, for its users and for its downstream
distributors to only use free sound files.

The new message sound file wasn’t even sourced, this makes it even more
problematic as the author can’t be contacted at all.


sualko:
I doubt that the ringtone has the required threshold of originality and can therefore be used at least from my point of view.

If I remember correctly the incoming message sound file was public domain, because otherwise I had added a license note.

If you find music with cc or something similar it would be great, but to use the same sound file for all notifications is no good idea. But even if you find a sound track which claims to be free to use, you can't be sure that the website is the creator. So there is always some kind of uncertainty if you distribute something which you didn't create by your own.

sualko:
How about this?

https://freesound.org/people/FoolBoyMedia/sounds/234524/
https://freesound.org/people/pogmothoin/sounds/351284/

PR #869, opened 2020-04-15
Update uk-UA.json

ostasevych:
Updated translation

PR #870, opened 2020-04-15
Update uk-UA.json

ostasevych:
Updated Ukrainian translation


sualko:
Thanks for your contribution. What's the difference to #869?

Btw it's easier to make those changes at our translation website: https://webtranslateit.com/en/projects/10365-JSXC/

ostasevych:
I found that several translations are not fitting well in the chat window with the length of symbols, so I just shortened them. No significant improvements. Besides, have you thought about using some web-based translation system, eg crowding, which is used by Pix-Art, eg?

sualko:
> I found that several translations are not fitting well in the chat window with the length of symbols, so I just shortened them. No significant improvements.

But what is the difference to your other pull request? See #869 

> Besides, have you thought about using some web-based translation system, eg crowding, which is used by Pix-Art, eg?

Jep. That was the link I posted earlier: https://webtranslateit.com/en/projects/10365-JSXC/

ostasevych:
Ah, yes! With this quarantine my memory became weak. Please, ignore pull request, I made changes via weblate. Sorry for confusing you.

PR #889, opened 2020-05-08
feat: xmpp ping (server-to-client) plugin

niryas:
A simple plugin implementation of Server-to-Client pings,
as part of the XMPP Ping Protocol Extension (_XEP-0199_).
A reply will be sent only to stanzas originating from the server.
Otherwise, an error will be sent.

Per _XEP-0199_, the "from" attribute should be the server domain.
Requests with account's bare jid as "from" would be allowed,
as some server implementations send pings this way (Ejabberd, e.g.).

Needed this for our fork (we do not need client-to-client or client-to-server pings), so here it is if anyone else needs a similar feature. :)

**Version**: 0.1
**See**: https://xmpp.org/extensions/xep-0199.html


sualko:
Thanks for your contribution :clap: . Will merge it after 4.1 is released.

niryas:
Let me know if there are any suggestions for changes.

And thank you for the great work on JSXC! :)

src/plugins/PingPlugin.ts
```diff
@@ -0,0 +1,95 @@
...
+            name: 'XMPP Ping',
+            version: '0.1',
+         }]
+      }
+   }
+
+   constructor(pluginAPI: PluginAPI) {
+      super(MIN_VERSION, MAX_VERSION, pluginAPI);
+
+      this.pluginAPI.addFeature(Namespace.get(PING));
```
sualko, line 55: I guess this should be `'PING'`

sualko:
Thanks again for your contribution. I merged your changes by hand.

PR #944, opened 2020-10-01
feat: Added support for custom http headers

shivshankardayal:
Please review this. I am not an expert of JSXC but this seems good to me. Let me know if I missed anything.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "ts-loader": "^7.0.2",
     "ts-node": "^8.10.1",
     "tslib": "^1.11.2",
-    "typescript": "^3.5.3",
+    "typescript": "^3.9.7",
     "typescript-formatter": "^7.2.2",
     "webpack": "^4.43.0",
     "webpack-bundle-analyzer": "^3.7.0",
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     }
   },
   "devDependencies": {
+    "@aws-amplify/auth": "^3.4.6",
     "@commitlint/cli": "^8.1.0",
     "@commitlint/config-conventional": "^8.1.0",
     "@commitlint/travis-cli": "^8.1.0",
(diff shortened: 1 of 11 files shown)


sualko:
Thanks for your pr and sorry that you have to wait that long. I will make some modifications to the code and merge it afterwards. Thanks again for your contribution :tada:

shivshankardayal:
Actually, I found that this PR is not complete and I have written mode code. I will add those commits to this PR as well.

sualko:
To be honest, there are multiple issues with the code. In general, one pr should only fix one issue or add one feature. In this pr, there are style changes, new bugs and also the way how headers are added could be enhanced. If would be great if different strophe options could be passed (e.g. mechanisms). Probably the easiest way to do this would be to use a client option. Therefore add a new field to `OptionsDefault.ts` and call `Client.getOption()` in `ConnectionHelper.ts`. What do you think?

sualko:
@shivshankardayal thanks for continue working on this pr. I really appreciate your work. There is something probably wrong with your last commit.

shivshankardayal:
Yes, adding this to `OptionsDefault.ts` would be nice. I did not want to change the way options work, thought that it will have undesirable changes for others. But if we change `OptionsDefault.ts` it would be much simpler to pass custom http headers to Srophe.

shivshankardayal:
@sualko I would suggest to close this PR as `OptionsDefault.ts` route seems more logical to me.

sualko:
> it would be much simpler to pass custom http headers to Srophe.

I think so too, and I don't see any negative impact there. Maybe it would be nicer to pass the options along side the start call in order to allow custom options if you connect to multiple services, but I think it's not worth the trouble.

> I would suggest to close this PR

It depends if you like to continue on it. In this case you could just overwrite your changes and leave the pr therefore open.

shivshankardayal:
I will start fresh with headers in options and write totally new code. No point continuing with this code. It was just a straight, linear and stupid approach to the problem which I had at hand.

sualko:
> stupid approach

I would not call it stupid, just straight forward. 

> will start fresh with headers in options

That are great news. I'm looking forward for your new pr. If you have time, it would be awesome if you could implement it in a way that you could pass all strophe options available. If you need help, please don't hesitate to contact me.

PR #948, opened 2020-11-06
feat: Added image paste feature (resolves #856)

### LINKED ISSUE #856: Add image paste
**Is your feature request related to a problem? Please describe.**
The modern chats contain the functionality of pasting not only text, but the images. Unavailability of this function leads to additional steps to be undertaken by potential user and thus make the app less comfortable for chatting.

**Is your feature request related to a specific XEP?**
XEP-0363

**Describe the solution you'd like**
Add pasting the image directly in the window of a chat. When pasted it is automatically http-uploaded to the server and ready to be sent to an interlocutor or muc.

**Additional context**
Furthermore additional tools on resizing, turning image may be added.

git-it:
JSXC is great! I'd like to help JSXC stay awesome :) 

This merge hopefully resolves #856 via a paste event trigger from the chatWindow input. This event triggers a check if the clipboard contains an image. If so the clipboard is read in as a blob and processed using the existing methods for file uploading.

### DIFF
--- a/src/ui/ChatWindowFileTransferHandler.ts
+++ b/src/ui/ChatWindowFileTransferHandler.ts
@@ -3,10 +3,11 @@ import Attachment from '../Attachment'
 
 export default class FileTransferHandler {
    private handlerElement;
-   private inputElement;
+   private inputElement: JQuery<HTMLElement>;
 
    constructor(private chatWindow: ChatWindow) {
       this.handlerElement = this.chatWindow.getDom().find('.jsxc-file-transfer');
+      this.inputElement = this.chatWindow.getDom().find('.jsxc-message-input');
 
       this.handlerElement.on('click', this.showFileSelection);
 
--- a/src/ui/ChatWindowFileTransferHandler.ts
+++ b/src/ui/ChatWindowFileTransferHandler.ts
@@ -3,11 +3,14 @@ import Attachment from '../Attachment'
 
 export default class FileTransferHandler {
    private handlerElement;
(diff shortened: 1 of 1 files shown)


src/ui/ChatWindowFileTransferHandler.ts
```diff
@@ -3,12 +3,15 @@ import Attachment from '../Attachment'
...
 export default class FileTransferHandler {
    private handlerElement;
+   private inputElement;
 
    constructor(private chatWindow: ChatWindow) {
       this.handlerElement = this.chatWindow.getDom().find('.jsxc-file-transfer');
 
       this.handlerElement.on('click', this.showFileSelection);
 
+      this.inputElement.on('paste', this.pasteImageAsFile);
```
sualko, line 13: Where gets the inputElement defined? At this point it should be undefined.

sualko (changes requested):
Thanks for your contribution and the kind words. It's nice to see that someone is enjoying jsxc.

src/ui/ChatWindowFileTransferHandler.ts
git-it: you are absolutely correct. 
Fixed this with commit 19c37a4719e688ed6a17fb96b98da3ff00444543. Line 10 gets the element from the dom (using the same method as done in chatWindow).
Thanks for your patience and guidance

sualko:
Thanks again for this nice feature :tada: :+1: Great work!

PR #954, opened 2020-12-09
Fix message duplicate issue when load mam

anvyzhang:
Issue is:
* User A and user B chat with each other, and a lot of messages archived to server
* User A use a new PC or web browser, or clear browser storage
* User B send user A message like "Hello"
* User A received message from user B "Hello", and click "Load history message"
* User A now received message "Hello" twice

Solve:
When get archived message, the query param "endDate" can be earlier than last message by 1 second, and wont miss any message.


sualko:
Did you close this pr intentionally?

anvyzhang:
Need to optimize that, and more test

PR #964, opened 2020-12-19
add menu to add/delete members from memberlist 

mightymop:
(xep-0045 9.5 Modifying the member list)


mightymop:
sorry wrong branch :D

PR #967, opened 2020-12-19
open help in new tab/window

mightymop:
to prevent connection loss if help is opened in same tab

### DIFF
--- a/src/ui/Roster.ts
+++ b/src/ui/Roster.ts
@@ -231,7 +231,7 @@ export default class Roster {
       }
    }
 
-   public addMenuEntry(options: { id: string, handler: (ev) => void, label: string | JQuery<HTMLElement>, icon?: string, offlineAvailable?: boolean }) {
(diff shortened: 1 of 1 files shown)


src/ui/Roster.ts
```diff
@@ -312,7 +312,8 @@ export default class Roster {
          this.addMenuEntry({
             id: 'online-help',
             handler(ev) {
-               window.location = onlineHelpUrl;
+               let win = window.open(onlineHelpUrl, '_blank');
```
sualko, line 315: This is likely to be blocked by popup blockers and noreferrer/noopener should be set. Maybe it's easier to make a fake click on the a label or something like that.

sualko (changes requested):

src/ui/Roster.ts
mightymop: fake click with shift ? so that it will open in a new window/tab?

mightymop: what about this:

let newTab = window.open();
newTab.location.href = onlineHelpUrl;
newTab.focus();

sualko (approved):

PR #968, opened 2020-12-19
increase windowsize - width from 250 to 320

mightymop:
Because the lock icon was not visible with 250px width, 320px is fine now

### DIFF
--- a/scss/main.scss
+++ b/scss/main.scss
@@ -13,7 +13,7 @@
 //fonts
 $font-sans: Arial, sans-serif;
 
-$default-window-width: 250px;
+$default-window-width: 320px;
 
 @import "partials/avatar";
 @import "partials/button";


sualko:
What do you mean with "the lock icon was not visible"? Sounds like a different issue, because I have no issue with the lock icon. Nonetheless on big screens this could definitely be increased.

image

mightymop:
> 
> 
> What do you mean with "the lock icon was not visible"? Sounds like a different issue, because I have no issue with the lock icon. Nonetheless on big screens this could definitely be increased.
> 
> image

hmm strange... because at my screens i only get the lock icon if i manually increase size of the chatwindow a little bit

sualko:
> hmm strange... because at my screens i only get the lock icon if i manually increase size of the chatwindow a little bit

Can you provide more information like browser, os and so on in order to fix this?

> by the way is there a full screen version planned?

We already provide a full screen version. Just add the classes `jsxc-fullscreen jsxc-two-columns` to your body. See the example for more details.

mightymop:
> 
> 
> > hmm strange... because at my screens i only get the lock icon if i manually increase size of the chatwindow a little bit
> 
> Can you provide more information like browser, os and so on in order to fix this?
> 
> > by the way is there a full screen version planned?
> 
> We already provide a full screen version. Just add the classes `jsxc-fullscreen jsxc-two-columns` to your body. See the example for more details.

Details: 

Browser: Firefox 84.0.1 (64-Bit)
OS: Windows 10 x64

Screenshots:
firefox1
firefox2

the lock will appear if i resize the chatwindow a little bit greater...

PR #969, opened 2020-12-20
XEP-0055 - Add User Search

mightymop:
[XEP-0055: Jabber Search](https://xmpp.org/extensions/xep-0055.html)

### DIFF
--- a/scss/partials/_dialog.scss
+++ b/scss/partials/_dialog.scss
@@ -224,3 +224,33 @@
     align-self: flex-start;
   }
 }
+
+.jsxc-content {
+  h1 {
+    font-size: 1.5em;
+    margin-top: 3em;
+  }
+}
+
+.jsxc-searchresults-table {
+  max-height: 600px;
+  overflow: auto;
+  width: 100%;
+
+  thead {
+    background-color: var(--jsxc-color-background-darker);
+  }
+
+  tr:nth-child(even) {
+    background-color: var(--jsxc-color-background-dark);
+  }
+
+  th {
+    text-align: center;
+  }
+
+  td {
+    padding: 5px;
+    text-align: center;
+  }
+}
--- a/src/connection/Form.ts
+++ b/src/connection/Form.ts
@@ -46,7 +46,6 @@ export default class Form {
          let itemFieldElements = $(itemElement).find('>field');
          return itemFieldElements.get().map((itemFieldElement) => ItemField.fromXML($(itemFieldElement)));
       });
-      //@TODO toHTML for reported and item elements
 
       return new Form(type, fields, instructions, title, reportedFields, items);
    }
@@ -74,7 +73,7 @@ export default class Form {
       return new Form('submit', fields);
    }
 
-   private constructor(private type: string, private fields: Field[], private instructions?: string, private title?: string, private reportedFields?: Field[], private items?: Field[][]) {
+   private constructor(private type: string, private fields: Field[], private instructions?: string, private title?: string, private reportedFields?: ReportField[], private items?: Field[][]) {
       if (this.ALLOWED_TYPES.indexOf(type) < 0) {
          throw new InvalidParameterError(`Form type not allowed! Instead of "${type}" try one of these: ${this.ALLOWED_TYPES.join(', ')}.`);
       }
@@ -95,10 +94,13 @@ export default class Form {
          }
 
          this.reportedFields.forEach((field, index) => {
-            if (this.reportedFields[index].getName() !== fields[index].getName()) {
+            //because the order of reported fields must not be equal with the fields order we have to search the field name in reportedFields array
+            //and cant simply use the same index here
+            let reportedFieldsArray = this.reportedFields.map(rfield => rfield.getName());
+            if (reportedFieldsArray.indexOf(fields[index].getName()) === -1) {
                throw new InvalidParameterError(`Item ${itemIndex} does not contain all "reported" fields.`);
             }
-         })
+         });
       });
    }
 
@@ -123,6 +125,14 @@ export default class Form {
    }
 
    public toHTML() {
+      if (this.type === 'form') {
+         return this.toHTMLForm();
+      } else if (this.type === 'result') {
+         return this.toHTMLResult();
+      }
+   }
+
+   private toHTMLForm() {
       let formElement = $('<form>');
       formElement.attr('data-type', this.type);
       formElement.attr('autocomplete', 'off');
@@ -147,6 +157,33 @@ export default class Form {
       }
 
       return formElement;
+   }
+
+   private toHTMLResult() {
+      let tableElement = $('<table>');
+      let tableHeader = $('<thead>');
+      let headerRow = $('<tr>');
+      let tableBody = $('<tbody>');
+
+      this.reportedFields.forEach(field => {
+         headerRow.append($('<th>').text(field.getName()));
+      });
+
+      tableHeader.append(headerRow).appendTo(tableElement);
+
+      this.items.forEach(fieldRow => {
+         let tableRow = $('<tr>');
+
+         fieldRow.forEach(field => {
+            $('<td>').text(field.getValues()[0]).appendTo(tableRow);
+         });
+
+         tableRow.appendTo(tableBody);
+      });
+
+      tableBody.appendTo(tableElement);
+
+      return tableElement;
    }
 
    public getValues(key: string): string[] {
--- a/src/ui/dialogs/contactsearch.ts
+++ b/src/ui/dialogs/contactsearch.ts
@@ -1,0 +1,244 @@
+import Dialog from '../Dialog'
+import Form from '../../connection/Form'
+import Log from '../../util/Log'
+import openContactDialog from './contact';
+import Translation from '@util/Translation';
+import Client from '../../Client'
+import Account from '../../Account'
+import JID from '../../JID'
+import { IJID } from '@src/JID.interface';
+import FormField from '@connection/FormField';
+
+export const CANCELED = 'canceled';
+
+const possibleSearchFields = ['first', 'last', 'nick', 'email'];
+
+let contactSearchTemplate = require('../../../template/contactsearch.hbs');
+
+export default function () {
+   let content = contactSearchTemplate();
+
+   let dialog = new Dialog(content);
+   dialog.open();
+
+   let accounts = Client.getAccountManager().getAccounts();
+   accounts.forEach(account => {
+      let accountoption = $('<option>').text(account.getJID().bare).val(account.getUid());
+
+      dialog.getDom().find('select[name="account"]').append(accountoption);
+   })
+
+   dialog.getDom().find('select[name="account"]').on('change', (ev) => {
+      let uid = $(ev.target).val().toString();
+
+      loadForm(Client.getAccountManager().getAccount(uid), dialog);
+   });
+
+   loadForm(accounts[0], dialog);
+}
+
+function loadForm(account: Account, dialog: Dialog) {
+   let searchService = account.getConnection().getSearchService();
+
+   getSearchService(account).then(async (services) => {
+      if (!services || services.length === 0) {
+         dialog.getDom().find('.jsxc-content').text(Translation.t('not_available'));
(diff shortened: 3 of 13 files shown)


scss/partials/_dialog.scss
```diff
@@ -216,3 +216,22 @@
...
 }
+
+.jsxc-searchresults-table {
+  max-height: 600px;
+  min-width: 400px;
+  overflow: scroll;
+  width: 400px;
+}
+
+.jsxc-searchresults-table thead tr th {
```
sualko, line 227: In Sass you can nest different rules to write less code.

```
.jsxc-searchresults-table {
  // your rules

  th { // th should be enough as selector
    // your th rules
  }
```

src/connection/Form.ts
```diff
@@ -91,14 +91,34 @@ export default class Form {
 
       this.items.forEach((fields, itemIndex) => {
          if (fields.length !== this.reportedFields.length) {
-            throw new InvalidParameterError(`Item ${itemIndex} does not contain all "reported" fields.`);
+             try{
+                 this.reportedFields.forEach((field, index) => {
+                    let i=0;
+                    let found=false;
+                    for (;i<fields.length;i++)
```
sualko, line 98: i

src/connection/Form.ts
```diff
@@ -91,14 +91,34 @@ export default class Form {
 
       this.items.forEach((fields, itemIndex) => {
          if (fields.length !== this.reportedFields.length) {
-            throw new InvalidParameterError(`Item ${itemIndex} does not contain all "reported" fields.`);
+             try{
+                 this.reportedFields.forEach((field, index) => {
```
sualko, line 95: In which case is the server returning more fields than reported fields?

src/connection/Form.ts
```diff
@@ -91,14 +91,34 @@ export default class Form {
...
+                    for (;i<fields.length;i++)
+                    {
+                        if (this.reportedFields[index].getName() === fields[i].getName()) {
+                            found=true;
+                            break;
+                        }
+                    }
+
+                    if (!found) {
+                       throw new InvalidParameterError(`Item ${itemIndex} does not contain all "reported" fields.`);
```
sualko, line 107: Why do we need the catch block, if we throw an error at this point?

src/ui/dialogs/contactsearch.ts
```diff
@@ -0,0 +1,253 @@
...
+      accounts: Client.getAccountManager().getAccounts().map(account => ({
+         uid: account.getUid(),
+         jid: account.getJID().bare,
+      }))
+    });
+
+    dialog = new Dialog(content);
+    dialog.open();
+
+    let accountId = <string> $('#jsxc-account').val();
```
sualko, line 29: I guess this is the same as `account.getUid()`.

src/ui/dialogs/contactsearch.ts
```diff
@@ -0,0 +1,253 @@
...
+        else
+        {
+            dialog.close();
+            alert(Translation.t('not_available'));
+        }
+    });
+
+}
+
+function getSearchServiceDomain(connection : IConnection, account : Account)
```
sualko, line 53: Should be converted to an async function.

src/ui/dialogs/contactsearch.ts
```diff
@@ -0,0 +1,253 @@
...
+    }
+    else
+        if ($(result).find('query').children('x').length>0)//Extended results
+        {
+            items=$(result).find('query').children('x').children('item');
+            if (items.length===0)
+            {
+                //NO RESULTS FOUND!
+                dialog.close();
+                alert(Translation.t('no_results'));
```
sualko, line 123: `alert` should not be used.

src/ui/dialogs/contactsearch.ts
```diff
@@ -0,0 +1,253 @@
...
+                return;
+            }
+            fields = $(result).find('query').find('reported').children('field');
+            let i=0;
+            for (;i<fields.length;i++)
```
sualko, line 128: i

src/ui/dialogs/contactsearch.ts
```diff
@@ -0,0 +1,253 @@
...
+                html+='</td>';
+            }
+        }
+
+        html+='</tr>';
+    }
+
+    html+='</tbody></table>';
+
+    dialog.getDom().find('#content').empty().append(html);
```
sualko, line 193: Needs a prefix and should not use an id.

sualko (changes requested):
Nice addition, but the code is hard to read. A separation between simple and extended results should help.

src/connection/Form.ts
mightymop: you are right ... maybe it was to late when i wrote the code...

src/connection/Form.ts
mightymop: ok it was to late :D

src/ui/dialogs/contactsearch.ts
mightymop: let accountId = <string> $('#jsxc-account').val();
let account = Client.getAccountManager().getAccount(accountId);

but need accountId to get account

scss/partials/_dialog.scss
mightymop: ok th and td now in the brackets of .jsxc-searchresults-table

src/ui/dialogs/contactsearch.ts
mightymop: done

src/ui/dialogs/contactsearch.ts
mightymop: ok now the dialog shows the error

src/ui/dialogs/contactsearch.ts
mightymop: done

mightymop:
> 
> 
> Nice addition, but the code is hard to read. A separation between simple and extended results should help.

tried to seperate some code now

src/ui/dialogs/contactsearch.ts
sualko: I guess it's the first account by default. Am I right?

src/ui/dialogs/contactsearch.ts
```diff
@@ -0,0 +1,278 @@
...
+let dialog: Dialog;
+
+export const CANCELED = 'canceled';
+
+let contactSearchTemplate = require('../../../template/contactsearch.hbs');
+
+export default function(username?: string) {
+
+    let content = contactSearchTemplate({
+      accounts: Client.getAccountManager().getAccounts().map(account => ({
```
sualko, line 20: The template is not using any variable.

src/ui/dialogs/contactsearch.ts
```diff
@@ -0,0 +1,278 @@
...
+    html+='</tr></thead><tbody class="jsxc-searchresults-table-body">';
+
+    for (let i=0;i<items.length;i++)
+    {
+        html+='<tr class="jsxc-searchresults-table-body-row '+(i%2===0?'jsxc-searchresults-odd':'jsxc-searchresults-even')+'"><td><input class="jsxc-searchresults-add" type="checkbox" name="jsxc-add-contact" value="'+i+'"/></td>';
+
+        html+='<td>';
+        if ($(items[i]).children('first').length>0)
+        {
+            html+=$(items[i]).children('first').text();
```
sualko, line 189: This allows XSS. We should not build any html from text.

src/ui/dialogs/contactsearch.ts
```diff
@@ -0,0 +1,278 @@
...
+        resolve(submitPromise);
+      });
+
+      formElement.find('.jsxc-js-close').click((ev) => {
+         ev.preventDefault();
+
+         dialog.close();
+
+         resolve(null);
+      });
```
sualko, line 276: The dialog close is looking for `.jsxc-js-close` and adding the close handler. So this can be removed.

src/ui/dialogs/contactsearch.ts
```diff
@@ -0,0 +1,278 @@
...
+   return new Promise((resolve, reject) => {
+        formElement.submit((ev) => {
+        ev.preventDefault();
+
+        let formval = Form.fromHTML(formElement.get(0));
+
+        let submitPromise = connection.getSearchService().executeSearchForm(jid,formval).then((stanza) => {
+            return stanza;
+        });
+        resolve(submitPromise);
```
sualko, line 267: `return connection.getSearchService().executeSearchForm(jid,formval)` is just the same.

sualko (changes requested):

src/ui/dialogs/contactsearch.ts
mightymop: we could also use: 
let account = Client.getAccountManager().getAccount()[0];

src/connection/Form.ts
```diff
@@ -90,15 +90,11 @@ export default class Form {
       }
 
       this.items.forEach((fields, itemIndex) => {
-         if (fields.length !== this.reportedFields.length) {
-            throw new InvalidParameterError(`Item ${itemIndex} does not contain all "reported" fields.`);
-         }
-
          this.reportedFields.forEach((field, index) => {
```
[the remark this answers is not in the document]
mightymop: Because you mentioned in an earlier version: "In which case is the server returning more fields than reported fields?"
And i think we could trust the servers implementation because we started with the extended form, too.

src/ui/dialogs/contactsearch.ts
mightymop: Just a C&P error... will remove it

src/ui/dialogs/contactsearch.ts
mightymop: changed to use jquery objects and methods now

src/ui/dialogs/contactsearch.ts
mightymop: without it does not close the dialog... maybe because its an dynamic form..

src/ui/dialogs/contactsearch.ts
mightymop: changed...

src/ui/dialogs/contactsearch.ts
mightymop: changed...

src/connection/Form.ts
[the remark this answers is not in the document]
sualko: We can't trust the server, because it's not always "our" server. If there is no issue with the check, what is the point in removing it?

sualko:
I refactored the code, so that it is more readable and fixed multiple issue regarding simple/extended search/results. Also the user is not directly added, but rather the add contact dialog is shown. This way the user is able to configure the name (alias).

Currently I have to possibility to test this code, so it would be great if you could debug this. We also need to add a account selection.

mightymop:
> 
> 
> I refactored the code, so that it is more readable and fixed multiple issue regarding simple/extended search/results. Also the user is not directly added, but rather the add contact dialog is shown. This way the user is able to configure the name (alias).
> 
> Currently I have to possibility to test this code, so it would be great if you could debug this. We also need to add a account selection.

ok with last commit i added an account selection and fixed some bugs 
I changed from checkboxes in resulttable to radio because of the new starting dialog. dialog chaining did not work.
If you change the account, the form will be reloaded, if new searchservices will have other forms...

template/contactsearch.hbs
```diff
@@ -1,5 +1,10 @@
...
+<div class="jsxc-account-choice">
+    <label class="col-sm-4 control-label" for="jsxc-account">{{t "Account"}}</label>
+    <div class="col-sm-8">
+     <select class="form-control" name="account" id="jsxc-account"></select>
+    </div>
+</div>
 <div class="jsxc-content">
     {{t "Loading"}}
 </div>
-<div class="jsxc-results"></div>
```
sualko: What was the reason to remove the result area? This way the user could change his search without closing, opening the dialog.

mightymop: re readded this... i did not thinkt about that ;)

sualko (approved):

PR #972, opened 2020-12-21
fix scrollbar in chatarea going outside

mightymop:
on some browsers the chatwindow scrollbar was not completly inside the chatbox

u see the scrollbar which is not completly inside and plz notice that the jsxc-bar on top has not the same size as the chatarea (there is a little white line on the right)

Firefox:

firefox2
firefox1

Chrome:

chrome1

### DIFF
--- a/scss/partials/_bar.scss
+++ b/scss/partials/_bar.scss
@@ -6,7 +6,7 @@
   height: 44px;
   padding: 0 3px;
   position: relative;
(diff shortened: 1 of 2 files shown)


sualko:
On the example page this doesn't occur (FF & Chrome), but I guess it's related to the used bootstrap style.

image

image

Issue #868, opened 2020-04-15
Add clickable full image preview to the chat window


ostasevych:
Please, add the functionality of full preview of the attached file when clicking on thumbnail in the chat,which will lead to popup window with the image.

PR #973, opened 2020-12-21
open links in new tab/window and load image urls

### LINKED ISSUE #911: Thumbnails only for images/ Url link for rest of file types
### Expected behavior
When files are received, the plateforme looks a lot more professional if they are displayed as thumbnails or at least icons instead of a long url link.

### Actual behavior

Except for images files received are displayed as url links including videos.

### Steps to reproduce the behavior
1.
2.
3.

### Environment
- **JSXC version:** 
- **Host system and version:** 
- **Browser vendor and version:** 
- **Any browser plugins enabled?** 
- **XMPP server vendor and version:** 
- **Is your XMPP server working with other clients as expected?** 
yes 

#### Javascript

```
```

#### JSXC

```
```

#### XMPP

```
```

#### Host

```
```

mightymop:
show image links as attachement and open links in new tab by setting target=_blank on links
fix #911
link1

### DIFF
--- a/src/plugins/httpUpload/HttpUploadPlugin.ts
+++ b/src/plugins/httpUpload/HttpUploadPlugin.ts
@@ -11,9 +11,12 @@ import { $iq } from '../../vendor/Strophe'
 import Translation from '../../util/Translation';
 import { IContact } from '@src/Contact.interface';
 import { IMessage } from '@src/Message.interface';
+import Client from '@src/Client';
 
 const MIN_VERSION = '4.0.0';
 const MAX_VERSION = '99.0.0';
+
+const IMAGE_SUFFIXES = ['jpeg', 'jpg', 'png', 'svg', 'gif'];
 
 export default class HttpUploadPlugin extends AbstractPlugin {
    public static getId(): string {
@@ -223,6 +226,21 @@ export default class HttpUploadPlugin extends AbstractPlugin {
       return Promise.resolve([message, xmlStanza]);
    }
 
+   private isTrustedDomain(url: URL): boolean {
+      let trustedDomains = Client.getOption<string[]>('trustedDomains', []);
+
+      return trustedDomains.filter(domain => {
+         let result = url.hostname === (domain);
+
+         if (!result && domain.indexOf('*.') > -1) {
+            let wildcardtestdomain = domain.substring(domain.lastIndexOf('*.') + 2);
+            result = url.hostname.endsWith(wildcardtestdomain);
+         }
+
+         return result;
+      }).length > 0;
+   }
+
    private extractAttachmentFromStanza = (contact: IContact, message: IMessage, stanza: Element): Promise<[IContact, IMessage, Element]> => {
       let element = $(stanza);
       let bodyElement = element.find('html body[xmlns="' + Strophe.NS.XHTML + '"]').first();
@@ -248,13 +266,48 @@ export default class HttpUploadPlugin extends AbstractPlugin {
                let attachment = new Attachment(name, mimeType, url);
                attachment.setThumbnailData(thumbnailData);
                attachment.setData(url);
-
                message.setAttachment(attachment);
                message.setPlaintextMessage(bodyElement.text());
             }
          }
+      } else {
+         this.processLinks(message);
       }
 
       return Promise.resolve([contact, message, stanza]);
-   };
+   }
+
+   private processLinks(message: IMessage) {
+      let plaintext = message.getPlaintextMessage();
+
+      if (!plaintext) {
+         return;
+      }
+
+      let pattern = new RegExp(/^(https?:\/\/[^\s]+)/);
+      let match = plaintext.match(pattern);
+
+      if (match) {
+         let url = match[0];
+         let extension = this.getFileExtensionFromUrl(url);
+
+         if (IMAGE_SUFFIXES.includes(extension)) {
+            let attachment = new Attachment('image', 'image/' + extension, url);
+            attachment.setData(url);
+
+            if (this.isTrustedDomain(new URL(url))) {
+               attachment.setThumbnailData(url);
+            } else {
+               attachment.setThumbnailData('../images/placeholder_image.svg');
+            }
+
+            message.setAttachment(attachment);
+            message.setPlaintextMessage(plaintext.replace(url, ''));
+         }
+      }
+   }
+
+   private getFileExtensionFromUrl(url: string): string {
+      return url.split(/[#?]/)[0].split('.').pop().trim().toLowerCase();
+   }
 }
--- a/src/ui/dialogs/settings.ts
+++ b/src/ui/dialogs/settings.ts
@@ -46,6 +46,14 @@ class ClientSection extends Section {
          this.getLanguageSelectionElement()
       ));
 
+      contentElement.append(new ListItem(
+         Translation.t('trusted_domains'),
+         Translation.t('one_domain_per_line'),
+         undefined,
+         undefined,
+         this.getTrustedDomainsElement()
+      ));
+
       return contentElement.getDOM();
    }
 
@@ -72,6 +80,21 @@ class ClientSection extends Section {
       if (element.find('[selected]').length === 0) {
          element.find('option:eq(0)').attr('selected', 'selected');
       }
+
+      return element;
+   }
+
+   private getTrustedDomainsElement(): JQuery {
+
+      let element = $('<textarea style="margin-left:10px;">');
+
+      element.on('change', () => {
+         let value = element.val().toString().split('\n').map(line => line.trim());
+         Client.setOption('trustedDomains',value ? value : undefined);
+      });
+
+      let trustedDomains = Client.getOption('trustedDomains', []);
+      element.val(trustedDomains.join('\n'));
 
       return element;
    }
--- a/images/placeholder_image.svg
+++ b/images/placeholder_image.svg
@@ -1,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   id="svg8"
+   version="1.1"
+   viewBox="0 0 26.458333 26.458334"
+   height="100"
+   width="100">
+  <defs
+     id="defs2" />
+  <metadata
+     id="metadata5">
+    <rdf:RDF>
+      <cc:Work
+         rdf:about="">
+        <dc:format>image/svg+xml</dc:format>
+        <dc:type
+           rdf:resource="http://purl.org/dc/dcmitype/StillImage" />
+        <dc:title></dc:title>
+      </cc:Work>
+    </rdf:RDF>
+  </metadata>
+  <g
+     transform="translate(0,-270.54165)"
+     id="layer1">
+    <rect
+       ry="0.58298022"
+       rx="0.58298022"
+       y="271.29044"
+       x="0.74882078"
+       height="24.960691"
+       width="24.960691"
+       id="rect815-5"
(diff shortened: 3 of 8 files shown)


src/plugins/httpUpload/HttpUploadPlugin.ts
```diff
@@ -158,7 +158,7 @@ export default class HttpUploadPlugin extends AbstractPlugin {
 
       let html = $('<div>').append(message.getHtmlMessage());
 
-      let linkElement = $('<a>');
+      let linkElement = $('<a target="_blank">');
```
sualko, line 161: We need to add `rel="noopener noreferrer"` for privacy and security protection.

src/plugins/httpUpload/HttpUploadPlugin.ts
```diff
@@ -227,6 +227,7 @@ export default class HttpUploadPlugin extends AbstractPlugin {
       let element = $(stanza);
       let bodyElement = element.find('html body[xmlns="' + Strophe.NS.XHTML + '"]').first();
       let dataElement = element.find('data[xmlns="urn:xmpp:bob"]');
+      let body = element.find('body');
```
sualko, line 230: That's to unspecific.

src/plugins/httpUpload/HttpUploadPlugin.ts
```diff
@@ -254,6 +255,32 @@ export default class HttpUploadPlugin extends AbstractPlugin {
             }
          }
       }
+      else
+          if (body.length===1 && !message.isEncrypted())
+          {
+              var link = body.text();
```
sualko, line 261: Url with text will not be detected.

src/plugins/httpUpload/HttpUploadPlugin.ts
```diff
@@ -254,6 +255,32 @@ export default class HttpUploadPlugin extends AbstractPlugin {
...
+                '((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.)+[a-z]{2,}|'+ // domain name
+                '((\\d{1,3}\\.){3}\\d{1,3}))'+ // OR ip (v4) address
+                '(\\:\\d+)?(\\/[-a-z\\d%_.~+]*)*'+ // port and path
+                '(\\?[;&a-z\\d%_.~+=-]*)?'+ // query string
+                '(\\#[-a-z\\d_]*)?$','i'); // fragment locator
+              if (!!pattern.test(link)&&link.lastIndexOf('.')>0)
+              {
+                   let suffix = link.substring(link.lastIndexOf('.')+1).toLowerCase();
+
+                   if (link.endsWith('.jpeg')||link.endsWith('.jpg')||link.endsWith('.png')||link.endsWith('.gif')||link.endsWith('.svg'))
```
sualko, line 273: could also be written as `['jpeg', ...].indexOf(suffix) > -1`.

sualko (changes requested):
This PR will introduce some privacy issues, because it automatically loads external content. We would need a "load image" button with some information about the implications and a configurable allow list for trusted domains.

src/plugins/httpUpload/HttpUploadPlugin.ts
mightymop: changed...

src/plugins/httpUpload/HttpUploadPlugin.ts
mightymop: is see jsxc will send urls with text in the following format:
url\ntext...
Now this will be parsed too.
Furthermore urls with example: http://domain:port/image.jpg?parameter... parameters will be parsed correctly too

src/plugins/httpUpload/HttpUploadPlugin.ts
mightymop: now using element.children('body') to get the body tag...

src/plugins/httpUpload/HttpUploadPlugin.ts
mightymop: added

src/plugins/httpUpload/HttpUploadPlugin.ts
sualko: There can also be multiple body children in XMPP. Just use the `message.getPlaintext()` method.

sualko (changes requested):
Without an opt-in option to load images, this pr can't be merged.

mightymop:
> 
> 
> Without an opt-in option to load images, this pr can't be merged.

So first showing a generic image as thumbnail with the ability to download is better? and furthermore a whitelist of allowed domains in settings!?

sualko:
> So first showing a generic image as thumbnail with the ability to download is better? and furthermore a whitelist of allowed domains in settings!?

That's right. Similar to email clients. Otherwise the user could be tracked.

mightymop:
> 
> 
> > So first showing a generic image as thumbnail with the ability to download is better? and furthermore a whitelist of allowed domains in settings!?
> 
> That's right. Similar to email clients. Otherwise the user could be tracked.

I think the intention of the Issuer was the he wants to see the images without doing sth. lazyness/comfort i guess, so the automatic download of images is wanted.
But i would change it in that way that if a domain is on whitelist the images are going to be downloaded automatically and if not then generic picture will be shown and the user has to click to download.

mightymop:
Settingsdialog now look like this:

src/plugins/httpUpload/HttpUploadPlugin.ts
```diff
@@ -246,14 +290,86 @@ export default class HttpUploadPlugin extends AbstractPlugin {
...
             }
          }
       }
+      else
+          if (body.length===1 && !message.isEncrypted())
+          {
+              let link = body.text();
+              let bodyText = link;
+              //Check if in normal body tag is text after URL to show
+              if (bodyText.split('\n').length>0)
```
sualko, line 332: Why are we requiring a newline to display the image?

src/ui/dialogs/settings.ts
```diff
@@ -37,15 +37,38 @@ class StartPage extends Page {
...
+         undefined,
+         undefined,
+         undefined,
+         undefined
+      ));
 
       contentElement.append(new ListItem(
          Translation.t('Language'),
-         Translation.t('After_changing_this_option_you_have_to_reload_the_page'),
+         undefined,
```
sualko, line 50: Why was this removed?

src/ui/dialogs/settings.ts
```diff
@@ -37,15 +37,38 @@ class StartPage extends Page {
...
          undefined,
          undefined,
          this.getLanguageSelectionElement()
       ));
 
+      contentElement.append(new ListItem(
+         Translation.t('trusted_domains'),
```
sualko, line 57: I think for the user it's not obvious what a trusted domain is. We need an explanation.

src/ui/dialogs/settings.ts
```diff
@@ -37,15 +37,38 @@ class StartPage extends Page {
...
+         Translation.t('one_domain_per_line'),
+         undefined,
+         undefined,
+         this.getTrustedDomainsElement()
+      ));
+
+	  contentElement.append(new ListItem(
```
sualko, line 64: Support for wildcards would be more flexibel. E.g. you could allow `*.data.foobar.com` and don't have to allow subdomains for every domain.

src/Attachment.ts
```diff
@@ -101,6 +101,11 @@ export default class Attachment {
       this.properties.set('thumbnail', thumbnail);
    }
 
+   public setGenericThumbnail()
```
sualko, line 104: We probably don't need this method, because we will always use the same image as "generic thumbnail". This means we can easily test if it's used or not.

sualko (changes requested):
This pr is a good starting point, but I think there are some things missing (which don't have to be implemented in this pr imo).

- loading thumbnail on user behalf
  - with explaination
  - with option to trust this domain in the future
- option to define a proxy to protect the users privacy

src/plugins/httpUpload/HttpUploadPlugin.ts
```diff
@@ -223,10 +226,51 @@ export default class HttpUploadPlugin extends AbstractPlugin {
       return Promise.resolve([message, xmlStanza]);
    }
 
+   private extractDomainFromUrl(url) {
```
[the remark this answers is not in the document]
mightymop: 🤦‍♂️ i need some hollidays again :D i know..

src/ui/dialogs/settings.ts
mightymop: just for design reasons but will revert it...

src/ui/dialogs/settings.ts
mightymop: added further infos

src/ui/dialogs/settings.ts
mightymop: wildcards added

src/Attachment.ts
mightymop: removed... and set generic image as default..
this would be overwritten if used otherwise with an other thumbnail...

src/plugins/httpUpload/HttpUploadPlugin.ts
```diff
@@ -246,14 +290,86 @@ export default class HttpUploadPlugin extends AbstractPlugin {
...
+
+               let domain = this.extractDomainFromUrl(url);
+               if (this.isTrustedDomain(domain))
+               {
+                  attachment.setThumbnailData(thumbnailData);
+               }
+               else
+               {
+                  attachment.setGenericThumbnail();
+               }
```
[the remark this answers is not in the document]
mightymop: ok

src/plugins/httpUpload/HttpUploadPlugin.ts
```diff
@@ -246,14 +290,86 @@ export default class HttpUploadPlugin extends AbstractPlugin {
...
+                '(\\?[;&a-z\\d%_.~+=-]*)?'+ // query string
+                '(\\#[-a-z\\d_]*)?$','i'); // fragment locator
+              if (!!pattern.test(link)&&link.lastIndexOf('.')>0)
+              {
+                   if (link.indexOf('?') > -1)
+                   {
+                       link = link.substring(0,link.lastIndexOf('?'));
+                   }
+
+                   let suffix = link.substring(link.lastIndexOf('.')+1).toLowerCase();
```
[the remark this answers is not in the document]
mightymop: using now

src/plugins/httpUpload/HttpUploadPlugin.ts
```diff
@@ -246,14 +290,86 @@ export default class HttpUploadPlugin extends AbstractPlugin {
...
+              let bodyText = link;
+              //Check if in normal body tag is text after URL to show
+              if (bodyText.split('\n').length>0)
+              {
+                  let bodyTextArr = body.text().split('\n');
+                  bodyTextArr[0]='';
+                  bodyText = bodyTextArr.map(item => item.trim()).join('\n').trim();
+              }
+
+              let pattern = new RegExp('^(https?:\\/\\/)?'+ // protocol
```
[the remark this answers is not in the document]
mightymop: ah you mean url ins textmessages embedded? hm I forgot this case... i only thought about uploading images .. but you are right... will change it..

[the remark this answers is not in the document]
mightymop: now extracting all URLs from messagetext and analyze them...
the first will become an attachment.. further links will become a clickable link...

src/plugins/httpUpload/HttpUploadPlugin.ts
mightymop: i process all links now...

sualko:
FYI I had to change a few things:

- links are already handled by jsxc, so no need to detect links twice (btw your implementation allowed xss)
- target blank was not needed at the second place, because the generated html is send to the peer and there is a very strict set of allowed attributes in xmpp
- only links at the beginning of the message are detected as attachment, because I don't want to alter the meaning of a message
- I added a placeholder image

One step further to create an awesome xmpp client. Thank you very much for your effort.

mightymop:
i think #868 could be closed too?

sualko:
> i think #868 could be closed too?

I think it's more about loading the image inside the ui/dialog. At least we should probably remove the `download` attribute from image links, so that the browser is not showing the download dialog.

PR #975, opened 2020-12-21
fix omemo key loading from s2s servers

mightymop:
If a omemo session in a chatbox is initiated by the user, the keys and devicelist are now loaded from a pep service on another server (server 2 server connection) too.

### DIFF
--- a/src/plugins/omemo/Plugin.ts
+++ b/src/plugins/omemo/Plugin.ts
@@ -68,9 +68,10 @@ export default class OMEMOPlugin extends EncryptionPlugin {
    }
 
    private openDeviceDialog = async (chatWindow: ChatWindow) => {
-      await this.getOmemo().prepare();
-
       let peerContact = chatWindow.getContact();
+
+      await this.getOmemo().prepare();
+      await this.refreshDeviceList(peerContact);
 
       return OmemoDevicesDialog(peerContact, this.getOmemo());
    }
@@ -87,7 +88,9 @@ export default class OMEMOPlugin extends EncryptionPlugin {
 
       return this.getOmemo().prepare().then(async () => {
          if (!this.getOmemo().isSupported(contact)) {
-            throw new Error(Translation.t('Your_contact_does_not_support_OMEMO'));
+            if (!this.refreshDeviceList(contact)) {
+               throw new Error(Translation.t('Your_contact_does_not_support_OMEMO'));
+            }
          }
 
          if (!this.getOmemo().isTrusted(contact) && !await this.getOmemo().trustOnFirstUse(contact)) {
@@ -98,6 +101,15 @@ export default class OMEMOPlugin extends EncryptionPlugin {
 
          OMEMOPlugin.updateEncryptionState(contact, trust);
       });
+   }
+
+   private async refreshDeviceList(contact: IContact) {
+      let pepService = this.pluginAPI.getConnection().getPEPService();
+      let stanza = await pepService.retrieveItems(NS_DEVICELIST, contact.getJid().bare);
+
+      this.onDeviceListUpdate(stanza);
+
+      return this.getOmemo().isSupported(contact);
    }
 
    private onDeviceListUpdate = (stanza) => {


src/plugins/omemo/Plugin.ts
```diff
@@ -100,6 +103,42 @@ export default class OMEMOPlugin extends EncryptionPlugin {
       });
    }
 
+   private getDeviceListFromContact(contact: IContact)
+   {
+      let connection = this.pluginAPI.getConnection();
+      let _self = this;
+      return connection.getPEPService().retrieveItems('eu.siacs.conversations.axolotl.devicelist',contact.getJid().bare).then(function(data) {
+
+            return new Promise(function(resolve, reject) {
```
sualko, line 112: No need for `_self` with arrow functions.

src/plugins/omemo/Plugin.ts
```diff
@@ -100,6 +103,42 @@ export default class OMEMOPlugin extends EncryptionPlugin {
...
+                if (_self.onDeviceListUpdate(data))
+                {
+                    const promises = [];
+                    let arr = $(data).find("device");
+                    let i=0;
+                    for (; i < arr.length; i++) {
+                       promises.push(connection.getPEPService().retrieveItems('eu.siacs.conversations.axolotl.bundles:'+$(arr[i]).attr('id'),contact.getJid().bare).then(function(data) {
```
sualko, line 119: No need to query all bundles. If a contact has a device list this should be enough.

src/plugins/omemo/Plugin.ts
```diff
@@ -100,6 +103,42 @@ export default class OMEMOPlugin extends EncryptionPlugin {
...
+                            return new Promise(function(resolve, reject) {
```
sualko, line 120: No need to return a new promise.

src/plugins/omemo/Plugin.ts
```diff
@@ -100,6 +103,42 @@ export default class OMEMOPlugin extends EncryptionPlugin {
...
+                       .catch((e) => {
+                           console.log(e);
+                           reject();
+                    });
+                    resolve(true);
+                }
+                else
+                {
+                    reject();
+                }
```
sualko, line 137: Should be simplified.

sualko (changes requested):
Looks like a workaround for non working pep notifications. In normal circumstances, those chances should not be needed.

mightymop:
> 
> 
> Looks like a workaround for non working pep notifications. In normal circumstances, those chances should not be needed.

You are right that normally it should not be needed... but i often get those errors like "Your contact does not support Omemo only because the keys where not received...". But before throwing such an error i like to check again to be sure. Especially on mobile devices when there were network changes outside and stream management is not available to detect package loss

src/plugins/omemo/Plugin.ts
mightymop: I tried withouth receiving the bundles only the key, then i got this error that my contact does not support omemo.
Then i notice in the omemo device/key dialog that the keys are not loaded.
If i force to load the keys, then they are avaiable...

src/plugins/omemo/Plugin.ts
mightymop: i added some comments..

src/plugins/omemo/Plugin.ts
mightymop: changed

src/plugins/omemo/Plugin.ts
mightymop: changed to lamda notation and removed _self var

sualko:
> But before throwing such an error i like to check again to be sure

That's probably a good idea, but it should be enough to check the availability of the device list. The rest should be handled by the plugin. Otherwise the bundles would be requested twice. Also please use the already provided namespace constants and so on.

> if the contact is member of another server / domain, will pep notifications delivered to federated servers?

Usually this should work, but if you use OpenFire I can't give any guaranties.

mightymop:
ok i removed the code for bundlereceiving...
maybe we should also load the devicelist if we open the omemo dialog... 
i opened before and there were no bundles until i enabled the lock icon and opened the dialog again...

sualko:
> maybe we should also load the devicelist if we open the omemo dialog...

I think that's a good idea.

sualko (approved):

sualko:
You `getDeviceListFromContact` always returned true and therefore every contact "supported" omemo :joy: 

I removed also the duplicated code in the dialog, but it should work now.

mightymop:
> You `getDeviceListFromContact` always returned true and therefore every contact "supported" omemo 
> 
> I removed also the duplicated code in the dialog, but it should work now.

Hoppla... thats why we review code  😇 but i am increasing skills hopefully...

PR #976, opened 2020-12-23
XEP-0084: PEP Avatar

### LINKED ISSUE #624: Display pubsub Avatars
Good day gentlemen.

As far as I can tell, there are two kinds of avatars possible with XMPP:
The vcard-based avatars and the pubsub-based avatars.

The very good messenger conversations is using the pubsub-based and jsxc the vcard-based and so the avatars from conversations will not be displayed in jsxc.

I think the best solution would be to request both of them and take the pubsub-based first and the vcard-based as an fallback solution.

If you agree me, I could implement this and make a pull request.

mightymop:
**PR was splitted...**
- [x] https://github.com/jsxc/jsxc/pull/997 (Avatarupload vCard) has to be merged before this 
- [x] load Avatars from PEP / URL 
- [ ] TODO: extend the Uploaddialog to upload PEP Avatars too

fix #624

I think there is space for further improvements... 
VCard Avatar and PEP Avatar have some difference workflows...
- Normaly pep Avatars are sent by server when user logs in. So the Avatars are received automatically as message events.
- vCard Avatars are pulled from Server...
- So the Problem is when both Plugins are activated it is possible that they overwrite each other, depends on message order !?

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,7 +20,6 @@ module.exports = {
         "plugin:@typescript-eslint/recommended",
         "plugin:@typescript-eslint/recommended-requiring-type-checking",
         "prettier",
-        "prettier/@typescript-eslint"
     ],
     "parser": "@typescript-eslint/parser",
     "parserOptions": {
--- a/package.json
+++ b/package.json
@@ -67,7 +67,6 @@
     "archiver": "^5.1.0",
     "buffer": "^6.0.3",
     "chai": "^4.2.0",
-    "clean-webpack-plugin": "^3.0.0",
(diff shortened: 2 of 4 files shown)


mightymop:
i am currently splitting this PR into
1. avatar upload feature and extend presence with update
2. xep 84 plugin

sualko:
Since I had to rewrite large parts of #997, this pr should be refactored as well.

mightymop:
> Since I had to rewrite large parts of #997, this pr should be refactored as well.

I will have a look on it when i have some time in the next few days

PR #977, opened 2020-12-23
add Tags to Roster Item

### LINKED ISSUE #77: Roster groups are ignored

mightymop:
tag1
tag2

Options:
- showTags: true = force Tag-text, false = circle with hover to show text
- blinkNewMessages: true = blinking avatar on new messages, false = no blinking

fix #77

this shows the roster groups as labels unter de statustext/lastmessages fields
multiple groups are joined in one line ... 
if there is to less place the string will be ellipsized and could be read by a mouseover action

### DIFF
--- a/src/ui/RosterItem.ts
+++ b/src/ui/RosterItem.ts
@@ -9,6 +9,7 @@ import Translation from '../util/Translation'
 import Client from '@src/Client';
 import Log from '@util/Log'
 import Color from '../util/Color'
+import Roster from '@ui/Roster'
 
 let rosterItemTemplate = require('../../template/roster-item.hbs')
 
@@ -31,8 +32,9 @@ export default class RosterItem {
       this.element.attr('data-presence', Presence[this.contact.getPresence()]);
       this.element.attr('data-subscription', this.contact.getSubscription());
       this.element.attr('data-date', this.contact.getLastMessageDate()?.toISOString());
+      this.element.attr('data-groups', this.contact.getGroups().join(',').toLowerCase());
 
-     this.appendTags();
+      this.appendTags();
 
       this.element.on('dragstart', (ev) => {
          (<any> ev.originalEvent).dataTransfer.setData('text/plain', contact.getJid().full);
@@ -151,11 +153,16 @@ export default class RosterItem {
    private appendTags() {
       let groups = this.contact.getGroups();
       let tagElements = groups.map(group => {
-         let element = $('<span>');
+         let element = $('<button>');
 
          element.text(group);
          element.addClass('jsxc-bar__tag');
          element.css('background-color', Color.generate(group));
+         element.on('click', ev => {
+            ev.preventDefault();
+
+            Roster.get().setFilter(group);
+         })
 
          return element;
       });
--- a/src/ui/RosterItem.ts
+++ b/src/ui/RosterItem.ts
@@ -8,6 +8,7 @@ import { IContact } from '../Contact.interface'
 import Translation from '../util/Translation'
 import Client from '@src/Client';
 import Log from '@util/Log'
+import Color from '../util/Color'
 
 let rosterItemTemplate = require('../../template/roster-item.hbs')
 
@@ -30,6 +31,8 @@ export default class RosterItem {
       this.element.attr('data-presence', Presence[this.contact.getPresence()]);
       this.element.attr('data-subscription', this.contact.getSubscription());
       this.element.attr('data-date', this.contact.getLastMessageDate()?.toISOString());
+
+     this.appendTags();
 
       this.element.on('dragstart', (ev) => {
          (<any> ev.originalEvent).dataTransfer.setData('text/plain', contact.getJid().full);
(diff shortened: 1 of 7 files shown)


src/ui/RosterItem.ts
```diff
@@ -132,15 +161,45 @@ export default class RosterItem {
...
+       let i=0
+       for (;i<groups.length;i++)
+       {
+           let indexgroup=RosterItem.allgroups.indexOf(groups[i]);
+           if (indexgroup===-1)
+           {
+               RosterItem.allgroups.push(groups[i]);
+               indexgroup=RosterItem.allgroups.length-1;
+               RosterItem.allgroupscolors.push(Color.generate(groups[i]));
+           }
```
sualko, line 196: `allgroups` and `allgroupscolors` should not be needed, because `Color.generate` always generates the same color for the same string.

src/ui/RosterItem.ts
```diff
@@ -132,15 +161,45 @@ export default class RosterItem {
...
+           let style='background-color: '+RosterItem.allgroupscolors[indexgroup]+';';
+           result.push('<span class="jsxc-bar-tags-item" style="'+style+'">'+groups[i]+'</span>');
+       }
+       return result.join('');
```
sualko, line 200: For security reasons I would prefer to generate some elements instead of a string.

src/ui/dialogs/vcard.ts
```diff
@@ -30,6 +32,18 @@ export default function(contact: IContact) {
    dialog = new Dialog(content);
    dialog.open();
 
+   if (Client.getOption('showTags',false))
```
sualko, line 35: I see no reason why groups should not always be displayed.

template/roster-item.hbs
```diff
@@ -4,6 +4,7 @@
    <div class="jsxc-bar__caption jsxc-grow">
       <div title="{{jid}}" class="jsxc-bar__caption__primary">{{name}}</div>
       <div title="{{lastMessage}}" class="jsxc-bar__caption__secondary">{{lastMessage}}</div>
+      <div title="{{tag}}" class="jsxc-bar-tags"></div>
```
sualko, line 7: To stay with the css BEM scheme, `jsxc-bar-tags` should be `jsxc-bar__tags`.

src/ui/RosterItem.ts
```diff
@@ -29,6 +43,21 @@ export default class RosterItem {
       this.element.attr('data-subscription', this.contact.getSubscription());
       this.element.attr('data-date', this.contact.getLastMessageDate()?.toISOString());
 
+      let groups = this.contact.getGroups();
+
+      if (Client.getOption(SHOW_TAGS,false)&&groups&&groups.length>0)
+      {
+          this.element.attr('data-groups', groups.join(';'));
```
sualko, line 50: Is the attribute actually used somewhere?

scss/partials/_roster.scss
```diff
@@ -363,3 +363,29 @@ li.jsxc-hide-offline {
...
+  z-index: 99;
+}
+
+.jsxc-bar--has-unread-msg-blink {
+  animation: blinker 1s linear infinite;
+}
+
+@keyframes blinker {
+  50% { opacity: 0; }
+}
```
sualko, line 391: I guess not everyone is a fan of a blinking icon.

template/roster-item.hbs
sualko, line 7: How does this work with the rename input?

sualko (changes requested):
Maybe we can also display last message and tags all at once. My idea would be to display only a circle by default and to enlarge the label with text on hover. What do you think?

src/ui/RosterItem.ts
mightymop: now a search over files dit not find it anywhere

scss/partials/_roster.scss
mightymop: maybe a flag for it? because i got some messages that a new message should be more present in roster, if a user did not see the notification

template/roster-item.hbs
mightymop: tag1
renaming of item is still ok

scss/partials/_roster.scss
mightymop: i added a new option blinkNewMessages: true = blinking, false = not

template/roster-item.hbs
mightymop: changed

src/ui/dialogs/vcard.ts
mightymop: OK i changed behavoir of showTags option to: true = force the text to be shown, false = circles with hover to show text

src/ui/RosterItem.ts
mightymop: changed

src/ui/RosterItem.ts
mightymop: removed

mightymop:
> 
> 
> Maybe we can also display last message and tags all at once. My idea would be to display only a circle by default and to enlarge the label with text on hover. What do you think?

ok as u can see in last commit i changed it to:

Options:
- showTags: true = force Tag-text, false = circle with hover to show text
- blinkNewMessages: true = blinking avatar on new messages, false = no blinking

sualko:
I changed the PR a bit:

- I removed the new option, because if you like to show the tag all the time, you can do it with some css. This is more coherent with the other options.
- I changed it back from a three to a two line item, because it was a bit squished.
- To distinguish tags from the unread notification bullet, tags are now displayed as squares.
- Blink option is also gone, because you can add it via css, but roster items with unread messages are more emphasized.
- You can filter for groups and set the filter if you click on a tag.
- On the vcard tags are only shown once now.

sualko (approved):

PR #978, opened 2020-12-24
Add dialog to enter status text

mightymop:


src/ui/Roster.ts
```diff
@@ -359,19 +360,26 @@ export default class Roster {
...
+            showSetExStatusDialog();
          }
-
-         if (oldPresence === Presence.offline && requestedPresence === Presence.online) {
-            let onUserRequestsToGoOnline = Client.getOption('onUserRequestsToGoOnline');
-
-            if (typeof onUserRequestsToGoOnline === 'function') {
-               onUserRequestsToGoOnline();
-            }
+         else
```
sualko, line 367: Returning inside the if clause would help to reduce indention and enhance the diff too.

src/ui/dialogs/exstatus.ts
```diff
@@ -0,0 +1,68 @@
...
+   dom.find('form').submit(onSubmit);
+
+   $('.jsxc-js-clear').on('click',removeStatus);
+}
+
+function onSubmit()
+{
+   let account = Client.getAccountManager().getAccounts()[0];
+   let statustext = dialog.getDom().find('#extended_status_text').val().toString();
+   let item = $build('status',{xmlns:'http://jabber.org/protocol/status'}).c('text').t(statustext).tree();
```
sualko, line 29: We should move this to a service.

src/ui/dialogs/exstatus.ts
```diff
@@ -0,0 +1,68 @@
...
+
+   dom.find('form').submit(onSubmit);
+
+   $('.jsxc-js-clear').on('click',removeStatus);
+}
+
+function onSubmit()
+{
+   let account = Client.getAccountManager().getAccounts()[0];
+   let statustext = dialog.getDom().find('#extended_status_text').val().toString();
```
sualko, line 28: Use e.g. form name instead of id.

src/ui/dialogs/exstatus.ts
```diff
@@ -0,0 +1,68 @@
+import Dialog from '../Dialog'
+import Client from '../../Client'
+//import JID from '../../JID'
```
sualko, line 3: Should be removed.

src/ui/dialogs/exstatus.ts
```diff
@@ -0,0 +1,68 @@
...
+
+function onSubmit()
+{
+   let account = Client.getAccountManager().getAccounts()[0];
+   let statustext = dialog.getDom().find('#extended_status_text').val().toString();
+   let item = $build('status',{xmlns:'http://jabber.org/protocol/status'}).c('text').t(statustext).tree();
+
+   account.getContact().setStatus(statustext);
+
+   account.getConnection().getPEPService().publish('http://jabber.org/protocol/status',item,'current').then(function(result) {
```
sualko, line 33: In which XEP is this defined?

template/roster.hbs
```diff
@@ -5,29 +5,33 @@
...
-        classes="jsxc-menu--pushup jsxc-grow jsxc-js-presence-menu"
-        button-label=(t "Offline")}}
-            <li data-presence="online" class="jsxc-online jsxc-offline-available">{{t "Online"}}</li>
-            <li data-presence="chat" class="jsxc-chat">{{t "Chatty"}}</li>
-            <li data-presence="away" class="jsxc-away">{{t "Away"}}</li>
-            <li data-presence="xa" class="jsxc-xa">{{t "Extended_away"}}</li>
-            <li data-presence="dnd" class="jsxc-dnd">{{t "dnd"}}</li>
-            <li data-presence="offline" class="jsxc-offline">{{t "Offline"}}</li>
-      {{/menu}}
+   <div class="jsxc-real-bottom">
```
sualko, line 8: Real bottom was probably not intentionally included in this pr.

sualko (changes requested):

src/ui/Roster.ts
mightymop: done

src/ui/dialogs/exstatus.ts
mightymop: ok i have moved the build code for the packets to pep service

src/ui/dialogs/exstatus.ts
mightymop: removed

src/ui/dialogs/exstatus.ts
mightymop: oh i now see it is not official... so we could remove it or not... up to you...
found it here: http://www.funkroom.net/xep/xep-status.html

template/roster.hbs
mightymop: right... C&P error ;D

src/ui/dialogs/exstatus.ts
mightymop: ok i have moved the build code for the packets to pep service

src/connection/services/PEP.ts
```diff
@@ -23,6 +23,16 @@ export default class PEP extends AbstractService {
       }
    }
 
+   public publisStatus(statustext:string)
```
sualko, line 26: Typo

src/ui/dialogs/exstatus.ts
```diff
@@ -0,0 +1,66 @@
...
+   });
+
+   $('.jsxc-js-save').on('click',sendStatus);
+   $('.jsxc-js-clear').on('click',removeStatus);
+}
+
+function sendStatus()
+{
+
+   let account = Client.getAccountManager().getAccounts()[0];
```
sualko, line 26: We should loop through all accounts

src/ui/dialogs/exstatus.ts
```diff
@@ -0,0 +1,66 @@
...
+let dialog: Dialog;
+
+export default function() {
+
+   let content = exstatusTemplate();
+
+   dialog = new Dialog(content);
+   let dom = dialog.open();
+
+   $(dom).ready(function(){
```
sualko, line 15: No need to wait

src/ui/dialogs/exstatus.ts
```diff
@@ -0,0 +1,66 @@
...
+   account.getConnection().getPEPService().publisStatus(statustext).then(function(result) {
+         dialog.close();
+   }).catch(function (msg){
+         dialog.close();
+   });
+}
+
+function getStatus()
+{
+     let account = Client.getAccountManager().getAccounts()[0];
```
sualko, line 43: We should check all accounts and than take the first one.

src/ui/dialogs/exstatus.ts
```diff
@@ -0,0 +1,66 @@
...
+     let jid = contact.getJid();
+     account.getConnection().getPEPService().retrieveItems('http://jabber.org/protocol/status',jid.bare).then(function(data) {
+         dialog.getDom().find('textarea[name="extended_status_text"]').val($(data).text());
+     });
+}
+
+function removeStatus(ev) {
+   ev.preventDefault();
+
+   let account = Client.getAccountManager().getAccounts()[0];
```
sualko, line 54: Loop

sualko (changes requested):
In my opinion we should not implement http://www.funkroom.net/xep/xep-status.html, because it's lacking crucial information. The biggest issue is, that it's not configuring a single item node. So we would probably end up with multiple old status messages. No big deal to fix it (see omemo), but I don't see a need for it either.

Otherwise looks good.

src/connection/services/PEP.ts
mightymop: removed...

src/ui/dialogs/exstatus.ts
mightymop: changed...

src/ui/dialogs/exstatus.ts
mightymop: added account choice

mightymop:
> 
> 
> In my opinion we should not implement http://www.funkroom.net/xep/xep-status.html, because it's lacking crucial information. The biggest issue is, that it's not configuring a single item node. So we would probably end up with multiple old status messages. No big deal to fix it (see omemo), but I don't see a need for it either.
> 
> Otherwise looks good.

i removed that unofficial xep..

src/ui/dialogs/exstatus.ts
mightymop: added account choice

src/ui/dialogs/exstatus.ts
mightymop: added account choice

sualko:
I merged it manually, because I changed some things:

- I removed the account and presence selection, because this would result in confusion. The roster bottom can only show one presence, so it would not be clear which account is using which presence.
- Status can now be added to every presence.
- The status is shown in the bottom part of the roster.

image

mightymop:
> 
> 
> I merged it manually, because I changed some things:
> 
>     * I removed the account and presence selection, because this would result in confusion. The roster bottom can only show one presence, so it would not be clear which account is using which presence.
> 
>     * Status can now be added to every presence.
> 
>     * The status is shown in the bottom part of the roster.
> 
> 
> image

very nice improvements...

PR #979, opened 2020-12-25
Add commands to muc chat window

### LINKED ISSUE #846: Kick/Ban users from group chat
Hello.
Are the commands of kick/ban implemented in group chat? It seems to me that these tools are not available, or I missed something?
https://xmpp.org/extensions/xep-0045.html#kick

mightymop:
To see all commands enter /help in a muc room
fix #846


scss/partials/_window.scss
```diff
@@ -801,3 +801,38 @@
...
+  display: none;
+  font-family: sans-serif;
+  overflow: hidden;
+  padding: 0;
+  position: absolute;
+  white-space: nowrap;
+  z-index: 1000;
+}
+
+.jsxc-custom-menu li {
```
sualko, line 819: We should taking advantage from the sass nesting possibilities here too.

src/ui/ChatWindow.ts
```diff
@@ -83,7 +84,65 @@ export default class ChatWindow {
       this.element.attr('data-presence', Presence[this.contact.getPresence()]);
       this.element.attr('data-subscription', this.contact.getSubscription());
 
-      this.getAccount().triggerChatWindowInitializedHook(this, contact);
+      this.getAccount().triggerChatWindowInitializedHook(this, contact);   this.element.find('div.jsxc-message-area').on('contextmenu',{ref:this},this.openContextMenu);
```
sualko, line 87: Newline missing

src/ui/ChatWindow.ts
```diff
@@ -83,7 +84,65 @@ export default class ChatWindow {
...
+
+      $(document).on('mousedown', function (e) {
```
sualko, line 89: Is this handler removed somewhere?

src/ui/ChatWindow.ts
```diff
@@ -83,7 +84,65 @@ export default class ChatWindow {
...
+          // If the clicked element is not the menu
+          // Hide it
+          if (!($(e.target).parents('.jsxc-custom-menu').length > 0)) {
+            $('.jsxc-custom-menu').hide(100);
```
sualko, line 93: Would hide all menus on the page.

src/ui/ChatWindow.ts
```diff
@@ -83,7 +84,65 @@ export default class ChatWindow {
...
+          }
+      });
+   }
+
+   private openContextMenu(event)
+   {
+        // Avoid the real one
+        event.preventDefault();
+        if (
+            $(event.target).closest('div.jsxc-chatmessage').hasClass('jsxc-probably_in')===false&&
```
sualko, line 103: Do we really need to compare it to false?

src/ui/ChatWindow.ts
```diff
@@ -405,6 +464,10 @@ export default class ChatWindow {
       }
    }
 
+   protected onInputCommand(message){
+       console.log('We are not in a muc, so ignore this call');
```
sualko, line 468: Use the `Log` class.

src/ui/MultiUserChatWindow.ts
```diff
@@ -5,12 +5,35 @@ import JID from '../JID'
 import AvatarSet from './AvatarSet'
 import showRoomConfigurationDialog from './dialogs/multiUserRoomConfiguration'
 import showMultiUserInviteDialog from './dialogs/multiUserInvite'
+import showMucHelp from './dialogs/muchelp';
+
+enum Command
+{
+    admin,// Zugehörigkeit des Benutzers zu Administrator ändern - wird über UI geregelt
```
sualko, line 12: Should be english.

src/ui/MultiUserChatWindow.ts
```diff
@@ -49,7 +79,74 @@ export default class MultiUserChatWindow extends ChatWindow {
       this.contact.registerMemberHook(this.contact.getNickname(), (data) => {
          this.updatePermissionAttributes(data);
       });
-      this.updatePermissionAttributes(this.contact.getMember(this.contact.getNickname()));
+
+      this.updatePermissionAttributes(this.contact.getMember(this.contact.getNickname()));   this.element.find('div.jsxc-message-area').on('contextmenu',{ref:this},this.openMUCContextMenu);
```
sualko, line 83: Newline missing.

src/ui/MultiUserChatWindow.ts
```diff
@@ -49,7 +79,74 @@ export default class MultiUserChatWindow extends ChatWindow {
...
+   }
+
+   private openMUCContextMenu(event)
+   {
+        event.data.ref.openContextMenu(event);
```
sualko, line 88: Use var for `event.data.ref`.

src/ui/MultiUserChatWindow.ts
```diff
@@ -49,7 +79,74 @@ export default class MultiUserChatWindow extends ChatWindow {
       this.contact.registerMemberHook(this.contact.getNickname(), (data) => {
          this.updatePermissionAttributes(data);
       });
-      this.updatePermissionAttributes(this.contact.getMember(this.contact.getNickname()));
+
+      this.updatePermissionAttributes(this.contact.getMember(this.contact.getNickname()));   this.element.find('div.jsxc-message-area').on('contextmenu',{ref:this},this.openMUCContextMenu);
+   }
+
+   private openMUCContextMenu(event)
```
sualko, line 86: The context menu could also be used in different scenarios like emoticons, or similar stuff.

src/ui/MultiUserChatWindow.ts
```diff
@@ -178,6 +275,173 @@ export default class MultiUserChatWindow extends ChatWindow {
       )
    }
 
+   protected onInputCommand(message)
+   {
+       console.log(this.contact);
```
sualko, line 280: Debug output should be removed.

src/ui/MultiUserChatWindow.ts
```diff
@@ -178,6 +275,173 @@ export default class MultiUserChatWindow extends ChatWindow {
...
+       let command = message.substring(1,message.indexOf(' ')!==-1?message.indexOf(' '):message.length).toLowerCase();
+       if (command === Command[Command.help])
```
sualko, line 282: I think if we use a similar structure like in https://github.com/jsxc/jsxc/blob/master/src/Notice.ts the code could be a lot more readable.

src/ui/dialogs/muchelp.ts
```diff
@@ -0,0 +1,14 @@
+import Dialog from '../Dialog'
+
+const muchelpTemplate = require('../../../template/muchelp.hbs');
+
+export default function() {
+   let content = muchelpTemplate({});
```
sualko, line 6: We should use camel case for the template file and maybe give it are more meaningful name.

sualko (changes requested):
That's a long requested feature and it really looks nice.

scss/partials/_window.scss
mightymop: if i move the li into the class, will the li then inherit from the class css fields?

src/ui/ChatWindow.ts
mightymop: The intention is to close the context menu if the user clicks outside of it

src/ui/ChatWindow.ts
mightymop: will "!" do it its the same... only shorthand... but i think we only need the contextmenu for incoming messages...

src/ui/MultiUserChatWindow.ts
mightymop: i named it openMUCContextMenu in the mucuserchatwindow class because the base class has its own openContextMenu method which is calld by the muc one... the muc method extends the base method... but dit not know how to call the super method ;D

src/ui/MultiUserChatWindow.ts
mightymop: as i said im not the typescript pro... so if you would merge and change it to your style then?

src/ui/dialogs/muchelp.ts
mightymop: done... now mucHelpDialog...

src/ui/MultiUserChatWindow.ts
mightymop: removed

src/ui/MultiUserChatWindow.ts
mightymop: done

src/ui/MultiUserChatWindow.ts
mightymop: NL added

src/ui/MultiUserChatWindow.ts
mightymop: translated..

src/ui/ChatWindow.ts
mightymop: using Log now

src/ui/ChatWindow.ts
mightymop: now removed on close method of the class

src/ui/ChatWindow.ts
mightymop: added NL

src/ui/ChatWindow.ts
mightymop: ok the mousedown handler is now bounded to the chatwindow only so i can remove the handler in the close method

scss/partials/_window.scss
sualko: No, it just "inherits" the selector.

```
a {
  // foo

  b {
    // bar
  }
}

// will be transformed to

a {
  // foo
}

a b {
  // bar
}
```

sualko:
Are you done with this pr? Because I would start to clean it up.

scss/partials/_window.scss
mightymop: ok changed the css

mightymop:
> 
> 
> Are you done with this pr? Because I would start to clean it up.

ok i finished my part here

sualko:
I generalized your code a bit more and moved all commands to a new plugin. Can you test my changes and verify that it is still working?

sualko (approved):

mightymop:
> 
> 
> I generalized your code a bit more and moved all commands to a new plugin. Can you test my changes and verify that it is still working?

The commands are working.

May be some things open:
- What about the translations in help dialog? Will they come back?
- And what about the context menu? I think you will add it again in another PR, won't u?
- I think we should send a presence to muc if we open the muc chatbox again. So that a kicked user will "rejoin"

sualko:
> What about the translations in help dialog? Will they come back?

I will add them via webtranslateit.com ater this pr is merged.

> And what about the context menu? I think you will add it again in another PR, won't u?

The option to quote a message will definitely be back and I have your context menu still in another branch, but I'm thinking about to use buttons for this actions similar to Signal desktop. The context menu introduces a new pattern (right click with menu) which the user is not used to. Maybe we can discuss this in another pr/issue.

> I think we should send a presence to muc if we open the muc chatbox again. So that a kicked user will "rejoin",
Now the muc has to be deleted and joined again with join dialog.

Maybe we have to show a message with a connect button or something like that. Not sure how we should handle that case exactly.

> Furthermore i have a strange behavior in the roster now. The MUC Rosteritem is present more than one time now... after kicking and joining... but if i will delete it, all of them are gone... so maybe a check on the add() is missing?

There is a check which should prevent double insertion, but we have to investigate that a bit further. Currently I try to merge everything...

PR #981, opened 2020-12-26
fix vcard parsing into list

mightymop:
old implementation did not parse correctly all data into the list for template creation

### DIFF
--- a/scss/partials/_dialog.scss
+++ b/scss/partials/_dialog.scss
@@ -10,6 +10,19 @@
   z-index: 999;
 }
 
+.jsxc-vcard-result {
+
+  margin-bottom:10px !important;
+
+  li > strong {
+    min-width: 150px;
+    display: inline-block;
+  }
+
+  li:nth-child(even) {background: var(--jsxc-color-background)}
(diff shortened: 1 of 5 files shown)


sualko:
Can you elaborate a bit more, which issue this pr is fixing?

mightymop:
> 
> 
> Can you elaborate a bit more, which issue this pr is fixing?

with your old implementation the conversion xml > json does not take all fields of the vcard... dont know why but there were some fields missed. mostly in deeper xml tags like adr and tel ...

mightymop:
i also added some css style to the vcard dialog now

sualko:
> does not take all fields of the vcard

Can you provide a copy of such an vcard?

Btw you don't have to squash your commits, because I do it once when merging and having multiple commits also helps to track your changes.

mightymop:
> 
> 
> > does not take all fields of the vcard
> 
> Can you provide a copy of such an vcard?
> 
> Btw you don't have to squash your commits, because I do it once when merging and having multiple commits also helps to track your changes.

ok.. on other repos they mostly want it squashed...

mightymop:
closed for PR: https://github.com/jsxc/jsxc/pull/984

PR #982, opened 2020-12-29
XEP-0191 Blocking Commands

### LINKED ISSUE #763: XEP-0191: Blocking Command
Support [XEP-0191: Blocking Command](https://xmpp.org/extensions/xep-0191.html) in order to achieve compliance to XEP-0412 (#756).

mightymop:
- Added menu to manage block list

- Todo: 
  - [ x] create handler to handle results from managed jids to enable/disable (un)blocked Buddys in Roster
     (Blocked buddys are colored in roster ind lined through)
  - [ x] disable the ability to write in allready opened Chatwindows if blocked     
     (Server will not route the message and should send an error back)

PS: if branch muccommands is merged we could add a new command like /block JID as quickaction for blocking and also in the context menu too

fix #763


sualko:
Just as a notice: If you add fix #number to the pr description, the pr is linked in the issue and the issue is automatically closed when the pr is merged.

mightymop:
> 
> 
> Just as a notice: If you add fix #number to the pr description, the pr is linked in the issue and the issue is automatically closed when the pr is merged.

ah nice to know

src/plugins/BlockingCommandPlugin.ts
```diff
@@ -0,0 +1,187 @@
...
+      pluginAPI.registerConnectionHook((status, condition) => {
+         if (status === Status.ATTACHED) {
+             // result in onReceiveBlockingItems
+            _self.getBlocklist();
+         }
+      });
+
+      pluginAPI.addAfterReceiveMessageProcessor(this.onMessageIncomingProcessor);
+
+      let roster = Roster.get();
```
sualko, line 66: This could maybe instantiate the roster to early. Not sure if this is used somewhere else in the code.

src/plugins/BlockingCommandPlugin.ts
```diff
@@ -0,0 +1,187 @@
...
+
+      pluginAPI.addFeature(NAMESPACE_BLOCKING_COMMAND);
+
+      connection.registerHandler(this.onReceiveBlockingItems, NAMESPACE_BLOCKING_COMMAND, 'iq');
+
+      let _self = this;
+      pluginAPI.registerConnectionHook((status, condition) => {
+         if (status === Status.ATTACHED) {
+             // result in onReceiveBlockingItems
+            _self.getBlocklist();
```
sualko, line 60: `_self` not needed, because arrow functions don't provide there own context.

src/plugins/BlockingCommandPlugin.ts
```diff
@@ -0,0 +1,187 @@
...
+
+      return this.pluginAPI.sendIQ(iq);
+   }
+
+
+   private onMessageIncomingProcessor = (contact: IContact, message: IMessage, stanza: Element): Promise<[IContact, IMessage, Element]> => {
+      let strfrom = $(stanza).attr('from');
+      if (strfrom.indexOf('/')>=0)
+      {
+          strfrom=strfrom.substring(0,strfrom.lastIndexOf('/'));
```
sualko, line 135: Use `JID` to get the bare jid.

src/ui/RosterItem.ts
```diff
@@ -40,6 +40,10 @@ export default class RosterItem {
       });
 
       this.element.click(function() {
+         if ($(this).hasClass('jsxc-rosteritem-disabled'))
+         {
+            return;
+         }
```
sualko, line 46: Is this really needed?

src/ui/dialogs/contactBlock.ts
```diff
@@ -0,0 +1,71 @@
...
+export default function() {
+   let content = contactBlockList({});
+
+   let dialog = new Dialog(content);
+
+   let account = Client.getAccountManager().getAccounts()[0];
+   let connection = account.getConnection();
+   let items;
+
+   BlockingCommandPlugin.getInstance().getBlocklist()
```
sualko, line 17: We need probably a `getPlugin` method in `PluginRepository` similar to `getEncryptionPlugin` for this.

src/ui/dialogs/contactBlock.ts
```diff
@@ -0,0 +1,71 @@
...
+          }
+    });
+}
+
+function initSubmit(dialog: Dialog, items, connection)
+{
+    dialog.getDom().find('form').on('submit', (ev) => {
+      ev.preventDefault();
+      let saveitems = dialog.getDom().find('textarea[name="jsxc-blocklist-textarea"]').val().toString().split('\n').map(item => item.trim())
+      saveitems=saveitems.filter(item => item.length>0);
```
sualko, line 40: We should test for valid jids.

sualko (changes requested):

src/plugins/BlockingCommandPlugin.ts
```diff
@@ -0,0 +1,187 @@
...
+         description: Translation.t('setting-vcard-avatar-enable'),
+         xeps: [{
+            id: 'XEP-0191',
+            name: 'Blocking Command',
+            version: '1.3',
+         }]
+      }
+   }
+
+   public static getInstance()
```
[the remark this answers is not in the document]
mightymop: How could i simply get the instance then?

src/plugins/BlockingCommandPlugin.ts
mightymop: changed...

src/ui/RosterItem.ts
mightymop: i add jsxc-rosteritem-disabled class in BlockingCommandPlugin if a user was blocked... because the user should not open a chatbox with a blocked user. It simply make no sense to write to a blocked user, if this user is not able to answer anymore..

src/ui/dialogs/contactBlock.ts
mightymop: I think testing for user jids will be enough?

src/ui/dialogs/contactBlock.ts
mightymop: Yes that would be nice... I also have this problem in other PRs too

src/plugins/BlockingCommandPlugin.ts
mightymop: changed...

src/plugins/BlockingCommandPlugin.ts
mightymop: Any suggestions?

src/ui/RosterItem.ts
sualko: I understand that, but you are hiding all disabled entries, so a user can't click that item.

src/ui/dialogs/contactBlock.ts
sualko: I think so too. You can checkout my member list code which is also testing for jids.

src/plugins/BlockingCommandPlugin.ts
sualko: No, I have to check that by my own. It was just a remark.

src/plugins/BlockingCommandPlugin.ts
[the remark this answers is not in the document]
sualko: Yep, you can run jsxc with multiple accounts.

> How could i simply get the instance then?

As mentioned below, we have to extend the plugin manager for that.

src/ui/dialogs/contactBlock.ts
mightymop: i have added the getPlugin() to PluginRepository ... is working fine...

src/plugins/BlockingCommandPlugin.ts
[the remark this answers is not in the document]
mightymop: i added a account selection and the suggested getPlugin()
now i get the instance and am able to work with it...

src/plugins/BlockingCommandPlugin.ts
mightymop: moved addMenuEntry for BlockDialog into connectionHook, so it will be…
d5686c2

… added on event CONNECTED (only once)

sualko (dismissed):

sualko:
According to the XEP-0191 you can also block resources and domains. My changes should reflect that and pull the list not on every page load and the ui state is synced across all tabs. Please check and maybe you have time to show a nice message if an account doesn't support this xep. Currently there will be just a "unknown error" message.

src/plugins/BlockingCommandPlugin.ts
```diff
@@ -146,10 +146,32 @@ export default class BlockingCommandPlugin extends AbstractPlugin {
...
          if (!newList.includes(jid)) {
             getElements(jid).removeClass('jsxc-blocked');
+
+            let contact = this.pluginAPI.getContact(new JID(jid));
+            if (contact)
+            {
+                let chatwindow =  contact.getChatWindow();
+                if (chatwindow)
+                {
+                   chatwindow.getElement().removeClass('jsxc-blocked');
```
sualko, line 156: This should not be needed, since `getElements` should include roster items and windows.

mightymop: i had the problem, that if i unblocked an user and opend the chatwindow again the title was still lined through

src/plugins/BlockingCommandPlugin.ts
```diff
@@ -146,10 +146,32 @@ export default class BlockingCommandPlugin extends AbstractPlugin {
...
+            }
          }
       });
 
-      blockedJids.forEach(jid => getElements(jid).addClass('jsxc-blocked'));
+      blockedJids.forEach((jid) => {
+          getElements(jid).addClass('jsxc-blocked');
+
+        let contact = this.pluginAPI.getContact(new JID(jid));
+        if (contact)
```
sualko, line 166: Btw this works only for full jids. To close all blocked jids we would need to go through the complete contact list and check every jid. Personally I think if a user blocks some jid, he can also close the chat window :-). Maybe the better idea would be to set the status (via `contact.setStatus()`) to blocked. This should be done in the `onBlocklistUpdate` since the `onBlocklistChanged` is executed in every tab.

mightymop: ok so should it still be done ?

sualko: If you like. Sure.

mightymop:
Hey... could we change the check in BlockingCommandPlugin.ts in method onMessageError like this?

```
private onMessageError = (stanza: string) => {
      if ($(stanza).find(`blocked[xmlns="${NAMESPACE_BLOCKING_COMMAND}:errors"]`).length === 0&&
          $(stanza).find(`not-acceptable[xmlns="urn:ietf:params:xml:ns:xmpp-stanzas"]`).length === 0 && 
          $(stanza).find(`service-unavailable[xmlns="urn:ietf:params:xml:ns:xmpp-stanzas"]`).length === 0 && 
          $(stanza).) {
         return true;
      }

     let fromval = $(stanza).attr('from');
     if (!fromval||fromval.length===0)
     {
        return true;
     }

     let from = new JID(fromval);
```

This would improve compatibilty with OF (yes i know...) but here is the ticket for this issue if you want to read it ...

> As for the error message: Openfire implements XEP-0016 Privacy Lists, and simply maps XEP-0191 on top of that. This is why you’re getting errors that do not correspond with XEP-0191, but with XEP-0016: https://xmpp.org/extensions/xep-0016.html#protocol-error. I’m not sure if this is easily fixable. Maybe we will need to track which XEP the user is using, and base the type of error to be generated on that. I’m kind of reluctant though, as that will get messy fast.

[https://igniterealtime.atlassian.net/browse/OF-2189](https://igniterealtime.atlassian.net/browse/OF-2189)

PR #984, opened 2020-12-30
Fix vcard - New PR

mightymop:
further CSS improvements and a little template change...

close #981
close #983


sualko:
The pr seams to be messed up someway. I try to cherry pick those changes and review if we can merge them. Currently I think it will have some issue on small screens.

PR #987, opened 2021-01-02
fix: use border-box

sualko:
@mightymop can you test if this fixes your out-of-window issue? Maybe also your hidden lock icon?

fix #972
fix #968


mightymop:
perfect ;) looks okay with lock icon and so on

PR #990, opened 2021-01-04
feat: add ability to read typeless messages

mightymop:
feat: Add ability to receive server broadcast messages
- Messages will be visible as notification + notification dialog

fix: add some missing translation vars
- "Open_window" and "Do_you_really_want_to_dismiss_all_notices"


sualko:
Broadcast messages are delivered with a [headline type](https://xmpp.org/rfcs/rfc6121.html#message-syntax-type) and are already handled by the `HeadlineMessageHandler`. Can you therefore explain your use case a bit more?

mightymop:
> 
> 
> Broadcast messages are delivered with a [headline type](https://xmpp.org/rfcs/rfc6121.html#message-syntax-type) and are already handled by the `HeadlineMessageHandler`. Can you therefore explain your use case a bit more?

Openfires broadcast from admin console simply looks like this:

`
<message from="mopsdom.de" xmlns="jabber:client"><body>TEST MESSAGE</body></message>
`

sualko:
> normal -- The message is a standalone message that is sent outside the context of a one-to-one conversation or groupchat, and to which it is expected that the recipient will reply.

Therefore Openfire is wrong and should rather be fixed than jsxc in my opinion.

> Type Attribute is optional so we nevertheless have to process a typeless message...

Only if we support normal messages, which we currently doesn't. I truly believe this is more an issue with Openfire than jsxc.

mightymop:
> 
> 
> > normal -- The message is a standalone message that is sent outside the context of a one-to-one conversation or groupchat, and to which it is expected that the recipient will reply.
> 
> Therefore Openfire is wrong and should rather be fixed than jsxc in my opinion.
> 
> > Type Attribute is optional so we nevertheless have to process a typeless message...
> 
> Only if we support normal messages, which we currently doesn't. I truly believe this is more an issue with Openfire than jsxc.



> 
> 
> > normal -- The message is a standalone message that is sent outside the context of a one-to-one conversation or groupchat, and to which it is expected that the recipient will reply.
> 
> Therefore Openfire is wrong and should rather be fixed than jsxc in my opinion.
> 
> > Type Attribute is optional so we nevertheless have to process a typeless message...
> 
> Only if we support normal messages, which we currently doesn't. I truly believe this is more an issue with Openfire than jsxc.

If i extend the notification dialog with the "typeless" message with a input for sending messages it will be the behavior of an normal message...

> normal -- The message is a standalone message that is sent outside the context of a one-to-one conversation or groupchat, and to which it is expected that the recipient will reply. Typically a receiving client will present a message of type "normal" in an interface that enables the recipient to reply, but without a conversation history.

sualko:
> Be able to read message is better than miss the messages completly.

That's right, but it seams that your pr (in openfire) is merged and therefore to add a workaround (with bad performance) is probably not needed.

mightymop:
> 
> 
> > Be able to read message is better than miss the messages completly.
> 
> That's right, but it seams that your pr (in openfire) is merged and therefore to add a workaround (with bad performance) is probably not needed.

This time their were very fast in merging my PRs @ igniterealtime :D I still have some open ones since spring 2020 ... -.-

mightymop:
PR closed because it is not needed anymore...

PR #991, opened 2021-01-05
fix: fix clear all notifications menuentry at bottom

mightymop:
- If there are more than ~ 25 unread notifications the
clear all menu item is not reachable, because the menu is noch scrollable.

### DIFF
--- a/scss/partials/_menu.scss
+++ b/scss/partials/_menu.scss
@@ -252,6 +252,15 @@
       left: 0;
       overflow: hidden;
       right: 0;
+
+      ul {
+        display: flex;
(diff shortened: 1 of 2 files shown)


sualko:
It's working fine, but I'm not sure if we should mess with the default behavior of lists. It would also work if we `prepend` new notices and `append` the delete all button in the roster class.

mightymop:
> 
> 
> It's working fine, but I'm not sure if we should mess with the default behavior of lists. It would also work if we `prepend` new notices and `append` the delete all button in the roster class.

You are right... i will change it

PR #974, opened 2020-12-21
Add more XEP infos to Settingsdialog

mightymop:
improved css for settings dialog
added tabs ... 

I took the XEPs from jsxc.org website, i did not proof the code for it...
So if all upcoming changes would be merged the xeps should be updated then...

PR #992, opened 2021-01-05
Enhance Settings

mightymop:
fix: link to corresponding XEP reference
feat: add more XEP Infos and collapsibles
fix: uncheck and disable checkbox for omemo plugin if libsignal is not available

close: #974

### DIFF
--- a/src/ui/DialogSection.ts
+++ b/src/ui/DialogSection.ts
@@ -3,8 +3,7 @@ import Navigation from './DialogNavigation'
 export default abstract class Section {
    private element: JQuery;
 
-   constructor(protected navigation: Navigation, private title?: string) {
-
+   constructor(protected navigation: Navigation, private title?: string, private isCollapsible: boolean = false) {
    }
 
    public getDOM(): JQuery {
@@ -19,9 +18,21 @@ export default abstract class Section {
       this.element = $('<section>');
 
       if (this.title) {
-         let legendElement = $('<h2>');
-         legendElement.text(this.title);
-         legendElement.appendTo(this.element);
+         if (this.isCollapsible) {
+            let button = $('<button>');
+
+            button.text(this.title);
+            button.addClass('jsxc-collapsible-settings-button');
+            button.on('click', (ev) => {
+               $(ev.target).toggleClass('jsxc-collapsible-settings-button-active');
+            });
+
+            button.appendTo(this.element);
+         } else {
+            let legendElement = $('<h2>');
+            legendElement.text(this.title);
+            legendElement.appendTo(this.element);
+         }
       }
 
       this.element.append(this.generateContentElement());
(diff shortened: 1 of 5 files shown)


src/ui/DialogSection.ts
```diff
@@ -2,9 +2,11 @@ import Navigation from './DialogNavigation'
 
 export default abstract class Section {
    private element: JQuery;
+   private button: JQuery;
+   private isCollapsible : Boolean;
 
-   constructor(protected navigation: Navigation, private title?: string) {
-
+   constructor(protected navigation: Navigation, protected collapsible?: Boolean, private title?: string) {
```
sualko, line 8: `Boolean` is a class, the correct type would be `boolean`.

src/ui/DialogSection.ts
```diff
@@ -2,9 +2,11 @@ import Navigation from './DialogNavigation'
...
+      this.isCollapsible=collapsible;
```
sualko, line 9: This is not needed if a access modifier is added to the constructor.

src/ui/DialogSection.ts
```diff
@@ -18,13 +20,35 @@ export default abstract class Section {
...
-         let legendElement = $('<h2>');
-         legendElement.text(this.title);
-         legendElement.appendTo(this.element);
+      if (this.isCollapsible)
+      {
+         this.button = $('<button class="jsxc-collapsible-settings-button">'+this.title+'</button>');
+         this.button.on('click',(e)=>{
+            $(e.target).toggleClass('jsxs-collapsible-settings-button-active');
+            let content = $(e.target).next();
+            if (content.css('display') === 'block') {
```
sualko, line 29: We can use css to hide/show the element. E.g. `.jsxs-collapsible-settings-button-active + div {display: block;}` or something should work.

src/ui/dialogs/settings.ts
```diff
@@ -79,7 +79,7 @@ class ClientSection extends Section {
 
 class AccountOverviewSection extends Section {
    constructor(navigation: Navigation) {
-      super(navigation, Translation.t('Accounts'));
+      super(navigation, false, Translation.t('Accounts'));
```
sualko, line 82: The title is a lot more common, so adding the new parameter `collapsible` to the end would simplify a lot.

sualko, line 82: Also adding a default parameter of `false` would better reflect the actual behaviour.

sualko (changes requested):
For me the bold buttons are breaking the design, but this is probably personal preference, but I'm wondering if we still need the inside scrollbar.

src/ui/DialogSection.ts
mightymop: changed..

src/ui/DialogSection.ts
mightymop: i refactored the jsxs-collapsible-settings-button class

src/ui/dialogs/settings.ts
mightymop: done

mightymop: done

PR #993, opened 2021-01-06
feat: add QR Codes in OmemodeviceDialog

mightymop:


src/ui/dialogs/omemoDevices.ts
```diff
@@ -39,15 +40,26 @@ class OmemoDeviceDialog {
...
+         this.insertDevices(contact.getJid().bare, peerDevices, identityManager, dom.find('.jsxc-omemo-peerdevices'));
+         this.insertDevices(contact.getAccount().getJID().bare, ownDevices, identityManager, dom.find('.jsxc-omemo-owndevices'));
          if (ownDevices.length > 1) {
             this.addCleanUpAction();
          }
+
       });
    }
 
+   private buildQRCode(jid: string, deviceid: string, key: string)
```
sualko, line 52: The device name is the bare jid, so we can just pass the device object.

src/ui/dialogs/omemoDevices.ts
```diff
@@ -39,15 +40,26 @@ class OmemoDeviceDialog {
...
+   {
+        let qr = new QRCode();
+        qr.setTypeNumber(5);
+        qr.setErrorCorrectLevel(ErrorCorrectLevel.L);
+        qr.addData('xmpp:'+jid+'?omemo-sid-'+deviceid+'='+key.split(' ').join(''));
```
sualko, line 57: Looking at the [commonly used url structure](https://github.com/psi-im/plugins/issues/59) (i will ask the omemo colleagues if there is any official documentation) you can verify all devices for one peer at once. This way the list will be a lot clearer.

sualko (changes requested):

src/ui/dialogs/omemoDevices.ts
mightymop: ok i changed it... :) combined the keys into one qr...

src/ui/dialogs/omemoDevices.ts
mightymop: changed

src/ui/dialogs/omemoDevices.ts
```diff
@@ -96,6 +136,16 @@ class OmemoDeviceDialog {
 
          listElement.append(element);
       }
+
+      let datauri = this.buildQRCode(devices[0].getAddress().getName(), fingerprints);
+      if (listElement.hasClass('jsxc-omemo-owndevices'))
+      {
+          $('.jsxc-own-device-qrcode-own').attr('src',datauri);
```
sualko, line 143: Could conflict with other elements on the website. Would be better to use `listElement.prev('.jsxc-qr-code').find('img')`.

sualko (changes requested):
Looks good to me. The only thing which is bothering me is that the qr library is not active maintained. Would be better to use https://www.npmjs.com/package/qrcode with @types/qrcode. I also don't think we have to calculate the error correction, because we can assume that the code will not be damaged on the screen.

src/ui/dialogs/omemoDevices.ts
mightymop: ok changed...

sualko (approved):

PR #995, opened 2021-01-08
feat: Add XEP-0202 Entity Time

mightymop:
Add [XEP-0202](https://xmpp.org/extensions/xep-0202.html) as a Plugin
At this time the plugin only reacts on querys from other clients/server.
May be we could improve it to query others for information and show theese on vcard dialog.

### DIFF
--- a/src/plugins/TimePlugin.ts
+++ b/src/plugins/TimePlugin.ts
@@ -1,7 +1,6 @@
 import { AbstractPlugin, IMetaData } from '../plugin/AbstractPlugin'
 import PluginAPI from '../plugin/PluginAPI'
 import Translation from '@util/Translation';
-import Contact from '../Contact'
 import JID from '../JID'
 
 const MIN_VERSION = '4.0.0';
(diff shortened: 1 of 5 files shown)


sualko:
Even after reading the xep, it's not clear to me what the purpose of this xep is. From a privacy point of view it seams like a bad idea to disclose the users timezone. At least it should answer only requests from contacts, or the own server.

mightymop:
> 
> 
> Even after reading the xep, it's not clear to me what the purpose of this xep is. From a privacy point of view it seams like a bad idea to disclose the users timezone. At least it should answer only requests from contacts, or the own server.

Because of that i implemented it as a plugin, so we ware able to disable that if not needed / wanted

src/plugins/TimePlugin.ts
```diff
@@ -0,0 +1,102 @@
...
+        let tojid = new JID($(stanza).attr('to'));
+        let type = $(stanza).attr('type');
+
+        if (type==='get'&&
+           (this.pluginAPI.getContact(fromjid)|| //only send to contacts
+            tojid.domain===fromjid.bare)) //or own domain server
+        {
+            let time = new Date();
+            let utc = time.toISOString();
+            let tzo = ((time.getTimezoneOffset()*-1)/60).toString();
```
sualko, line 95: timezone offset !== utc offset

The utc offset does not need to be a full hour.

PR #996, opened 2021-01-08
feat: XEP-0092 Software Version

mightymop:
add [XEP-0092](https://xmpp.org/extensions/xep-0092.html) as a Plugin
At this time the plugin only reacts on querys from other clients/server.
May be we could improve it to query others for information and show theese on vcard dialog.

### DIFF
--- a/src/plugins/VersionPlugin.ts
+++ b/src/plugins/VersionPlugin.ts
@@ -40,7 +40,7 @@ export default class VersionPlugin extends AbstractPlugin {
    }
 
    //query information from contact
-   public queryVersion(jid: JID): Promise<{}> {
+   public queryVersion(jid: JID): Promise<Element> {
       let iq = $iq({
          type: 'get',
          to: jid.full,
@@ -53,7 +53,7 @@ export default class VersionPlugin extends AbstractPlugin {
    }
 
    //send information response
-   private sendResponse(idstr: string, jid: string, name?: string, version?: string, os?: string): Promise<{}> {
+   private sendResponse(idstr: string, jid: string, name?: string, version?: string, os?: string): Promise<Element> {
       let iq = $iq({
          type: 'result',
          to: jid,
--- a/src/plugins/VersionPlugin.ts
+++ b/src/plugins/VersionPlugin.ts
@@ -1,0 +1,118 @@
+import { AbstractPlugin, IMetaData } from '../plugin/AbstractPlugin'
+import PluginAPI from '../plugin/PluginAPI'
+import Translation from '@util/Translation';
+import Client from '../Client'
+import JID from '../JID'
+
+const MIN_VERSION = '4.0.0';
+const MAX_VERSION = '99.0.0';
+
+const NAMESPACE_VERSION = 'jabber:iq:version'
+
+export default class VersionPlugin extends AbstractPlugin {
+
+   public static getId(): string {
+      return 'version';
+   }
+
+   public static getName(): string {
+      return 'Software Version';
+   }
+
+   public static getMetaData(): IMetaData {
+      return {
+         description: Translation.t('setting-version'),
+         xeps: [{
+            id: 'XEP-0092',
+            name: 'Software Version',
+            version: '1.1',
+         }]
+      }
+   }
+
+   constructor(pluginAPI: PluginAPI) {
+      super(MIN_VERSION, MAX_VERSION, pluginAPI);
+
+      pluginAPI.addFeature(NAMESPACE_VERSION);
+
+      let connection = pluginAPI.getConnection();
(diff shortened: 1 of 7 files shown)


src/plugins/VersionPlugin.ts
```diff
@@ -0,0 +1,93 @@
...
+
+      connection.registerHandler(this.onReceiveQuery, NAMESPACE_VERSION, 'iq');
+
+   }
+
+   //if name, version, os are empty it will be used as query for retrieve informations
+   public sendVersion(idstr: string, jid: string, name?: string, version?: string, os?: string): Promise<{}>
+   {
+      let iq = $iq({
+         type: (!name&&!version&&!os?'get':'result'),
```
sualko, line 47: Functionality and function name do not align.

src/plugins/VersionPlugin.ts
```diff
@@ -0,0 +1,93 @@
...
+         to: jid,
+         id: idstr,
```
sualko, line 49: Only required for responses, because strophe is generating one for requests.

src/plugins/VersionPlugin.ts
```diff
@@ -0,0 +1,93 @@
...
+      let connection = pluginAPI.getConnection();
+
+      pluginAPI.addFeature(NAMESPACE_VERSION);
+
+      connection.registerHandler(this.onReceiveQuery, NAMESPACE_VERSION, 'iq');
+
+   }
+
+   //if name, version, os are empty it will be used as query for retrieve informations
+   public sendVersion(idstr: string, jid: string, name?: string, version?: string, os?: string): Promise<{}>
```
sualko, line 44: Return type should be `Promise<Element>`.

src/plugins/VersionPlugin.ts
```diff
@@ -0,0 +1,93 @@
...
+      if (os)
+      {
+          iq.c('os').t(os).up();
+      }
+
+      return this.pluginAPI.sendIQ(iq);
+   }
+
+   private onReceiveQuery = (stanza) => {
+        let OSName='Unknown OS';
```
sualko, line 74: Would prefer to send an empty string, because this way other clients can correctly interpret the result.

src/plugins/VersionPlugin.ts
```diff
@@ -0,0 +1,93 @@
...
+
+        if (/android/i.test(userAgent)) {
+           OSName = 'Android';
+        }
+
+        if (/iPad|iPhone|iPod/.test(userAgent) && !window.MSStream) {
+            OSName = 'iOS';
+        }
+
+        this.sendVersion($(stanza).attr('id'),$(stanza).attr('from'),'JSXC',Client.getVersion(),OSName);
```
sualko, line 90: We should only respond to contacts.

sualko (changes requested):
Looks good, we just have to consider https://xmpp.org/extensions/xep-0092.html#security

src/plugins/VersionPlugin.ts
mightymop: How can i get the roster in a plugin?

src/plugins/VersionPlugin.ts
mightymop: changed...

src/plugins/VersionPlugin.ts
mightymop: ok

src/plugins/VersionPlugin.ts
mightymop: thats because i wanted to do two functionalities in one method. but i have two now...

src/plugins/VersionPlugin.ts
mightymop: changed..

src/plugins/VersionPlugin.ts
mightymop: i would go over Client.getAccountManager()... loop all accounts and so on?

sualko: Just try to get the contact from the plugin api (`api.getContact()`).

mightymop: thanks.. i should read the api source in future..

sualko: Thats too much :joy: No problem. It's a pleasure to help.

mightymop: changed...

mightymop:
> 
> 
> Looks good, we just have to consider https://xmpp.org/extensions/xep-0092.html#security

I added an option 'allowOSInfo' to allow sending info about OS or not

sualko (approved):

PR #997, opened 2021-01-09
feat: add dialog to upload vCard Avatar

### LINKED ISSUE #748: Avatar upload
Add ability to upload avatar in use with [XEP-0084: User Avatar](https://xmpp.org/extensions/xep-0084.html) and [XEP-0153: vCard-Based Avatars](https://xmpp.org/extensions/xep-0153.html). Maybe related to [XEP-0398: User Avatar to vCard-Based Avatars Conversion](https://xmpp.org/extensions/xep-0398.html) and requires probably #304.

### LINKED ISSUE #900: Click avatar to change status
### **Is your feature request related to a problem? Please describe.**

When I try to change Online status, I often click my avatar/icon. 

This behavior is expected in Nextcloud to access Nextcloud settings etc.

Until I clicked my name, I initially thought I couldn't change status. As it appears to be a link with the hand symbol... but nothing happens

### **Describe the solution you'd like**

Expand area to include the user avatar to change online status.

### LINKED ISSUE #985: Avatars not loading
### Expected behavior
Load Avatars on connect and reconnect

### Actual behavior
Even with a simple flat vCard my Avatars dont load.
I can see that getAvatar() in AvatarVCardPlugin loaded the vCard with the avatar image and calls resolve(...)
But then in getAvatar() of Contact class it fails because [, avatar] << avatar is undefined...
So no avatars are loaded, only the placeholders... neither my own and nor others in Roster...

### Environment
- **JSXC version:** 4.2.1
- **Host system and version:** self hostet
- **Browser vendor and version:** Firefox and Chrome latest
- **Any browser plugins enabled?** no
- **XMPP server vendor and version:** Openfire 4.6
- **Is your XMPP server working with other clients as expected?** yes

#### Javascript

``` 
throw new Error('No avatar available for ' + this.getId()); in Contact.getAvatar() is called...
```

#### JSXC

```
```

#### XMPP

```
```

#### Host

```
```

mightymop:
- extend Presence with x element for publish photo hash
- moved sha1 calculation methods to own file
- click on avatar in menu will open dialog

fix #900
fix #748
fix #985


mightymop:
I dont know if its allready in, but jsxc should load the clients own vcard after login. So the client is able to provide the hash of its own avatar then...

sualko:
> I dont know if its allready in, but jsxc should load the clients own vcard after login. So the client is able to provide the hash of its own avatar then...

That's right. It should be easily implementable in the plugin. There was just no time to do it.

One other think: Please only add translations to `en.json`, because we get a lot of merge conflicts and I also have issues to transfer other translations to webtranslate.it. Thanks.

mightymop:
> > I dont know if its allready in, but jsxc should load the clients own vcard after login. So the client is able to provide the hash of its own avatar then...
> 
> That's right. It should be easily implementable in the plugin. There was just no time to do it.

Ok

> One other think: Please only add translations to `en.json`, because we get a lot of merge conflicts and I also have issues to transfer other translations to webtranslate.it. Thanks.

Ok thanks for the info. I did not know whether you translate from german or english to other languages... so i provided both. So in future english only... 😇

src/connection/services/Vcard.ts
```diff
@@ -17,11 +18,129 @@ export default class Vcard extends AbstractService {
...
+   {
+      let getVcardIQ = $iq({
+         type: 'get',
+         to: jid.bare
+      }).c('vCard', {
+         xmlns: NS.get('VCARD')
+      });
+
+      //first get the actual vcard to merge image into
+      return this.sendIQ(getVcardIQ).then((stanza) => {
```
sualko, line 31: Would it not be easier to use `loadVcard`?

src/connection/services/Vcard.ts
```diff
@@ -17,11 +18,129 @@ export default class Vcard extends AbstractService {
...
+   private sendVCard(jid : IJID, avatar: string, vcard, getVcardIQ): Promise<any>
+   {
+       let sendVcardIQ = $iq({
+                 type: 'set',
+                 to: jid.bare
+              }).cnode(vcard[0]);
+
+        //second send the new vCard with avatar to server
+        return this.sendIQ(sendVcardIQ).then((result)=>{
+          if ($(result).attr('type')==='result')
```
sualko, line 91: Resolved promise should always from type result

src/connection/services/Vcard.ts
```diff
@@ -17,11 +18,129 @@ export default class Vcard extends AbstractService {
...
+                else
+                    return false;
+            });
+          }
+          else
+            return false;
+      });
+   }
+
+   private sendPresence(avatar, jid: IJID) {
```
sualko, line 125: Looks like duplicated code. Use `sendPresence` from `AbstractConnection` or even better trigger a resend, because we also want to include the status.

src/plugins/AvatarVCardPlugin.ts
```diff
@@ -60,6 +60,10 @@ export default class AvatarVCardPlugin extends AbstractPlugin {
             }
 
             let sha1OfAvatar = photo.text();
+            if (sha1OfAvatar!==null&&sha1OfAvatar.trim().length===0)
```
sualko, line 63: What was the reason for this check? Could also be written as `let sha1OfAvatar = photo.text()?.trim() || null;`

src/ui/AvatarSet.ts
```diff
@@ -37,6 +37,9 @@ export default class AvatarSet {
             let element = $(this);
 
             element.css('background-image', `url(${avatar.getData()})`);
+            element.css('background-size', 'contain');
+            element.css('background-repeat', 'no-repeat');
+            element.css('background-position', 'center');
```
sualko, line 42: Should be handled by css. Did you have issues with that?

src/ui/dialogs/avatarupload.ts
```diff
@@ -0,0 +1,177 @@
...
+   let content = avatarUploadTemplate();
+
+   dialog = new Dialog(content);
+   let dom = dialog.open();
+
+   let accounts = Client.getAccountManager().getAccounts();
+   for (let i=0;i<accounts.length;i++){
+       let accountoption = $('<option value="'+i+'">'+accounts[i].getUid()+'</option>');
+       dialog.getDom().find('select[name="account"]').append(accountoption);
+   }
```
sualko, line 26: I see the use cases that you like to have different avatars for your different accounts, but how can we visualize that in the roster?

src/ui/dialogs/avatarupload.ts
```diff
@@ -0,0 +1,177 @@
...
+
+   dialog.getDom().find('select[name="account"]').change((e)=>{
+       let val = Number($(e.target).val());
```
sualko, line 29: I think `parseInt(val, 10)` would be better.

src/ui/dialogs/avatarupload.ts
```diff
@@ -0,0 +1,177 @@
...
+       account=accounts[val];
+       loadAvatar(account,dialog);
+   });
+
+   console.log(mimetype);
```
sualko, line 34: debug output

src/ui/dialogs/avatarupload.ts
```diff
@@ -0,0 +1,177 @@
...
+}
+
+function handleFileSelect(file) {
+
+  if (!window.File || !window.FileReader || !window.FileList || !window.Blob) {
+    alert('The File APIs are not fully supported in this browser.');
+    return;
+  }
+
+  let fr = new FileReader();
```
sualko, line 68: I think the file reader ist not needed. Should be the same logic as attaching a file (`ChatWindowFileTransferHandler` and `Attachment`).

src/ui/dialogs/avatarupload.ts
```diff
@@ -0,0 +1,177 @@
...
+    mimetype=result.type;
+
+    $(img).attr('src', 'data:'+mimetype+';base64,'+base64data);
+    img.onload=null;
+    img.width=result.width;
+    img.height=result.height;
+    $(img).css('min-height', height);
+    $(img).css('min-width', width);
+    $(img).css('max-height', height);
+    $(img).css('max-width', width);
```
sualko, line 99: ?

src/ui/dialogs/avatarupload.ts
```diff
@@ -0,0 +1,177 @@
...
+    $(img).css('height', height);
+    $(img).css('width', width);
+    $(img).css('display','');
+  }
+  img.src = result.target.result;
+}
+
+function scaleDownTo9kb(img, start) : any
```
sualko, line 107: Maybe we can move this to a utility class and use the same logic as the attachment.

src/ui/dialogs/avatarupload.ts
```diff
@@ -0,0 +1,177 @@
...
+    base64data=dataurl.substring(dataurl.indexOf('base64,')+7);
+    let data=atob(base64data);
+
+    if (data.length>1024*8)
+    {
+        return scaleDownTo9kb(img, start-1);
+    }
+    else
+    {
+        return {'height':height,'width':width,'base64':base64data,'type':dataurl.substring(5,dataurl.indexOf(';',6)),size:data.length};
```
sualko, line 160: Looks prone for errors.

src/util/SHA1Hash.ts
```diff
@@ -0,0 +1,19 @@
+import * as sha1 from 'js-sha1'
+
+export default function calculateHash(data): string {
```
sualko, line 3: Should be renamed to something like `calculateHashFromBase64`

sualko (changes requested):

src/ui/AvatarSet.ts
mightymop: i had issues in chrome... chrome complained about the dataurl as invalid property when there was no background-size given...

src/plugins/AvatarVCardPlugin.ts
mightymop: i want to be sure that sha1OfAvatar is null for "no photo given" if there are some whitespaces in the phototag

src/connection/services/Vcard.ts
mightymop: ok

src/ui/dialogs/avatarupload.ts
mightymop: you mean the presence status bar on the bottom? good question... 
1. we could only show the first's account avatar there and serve the other avatars only for the contacts...
2. show a litte animation and switch between all avatars
3. show the first avatar and while hovering switch between others or show a little box with all others...
other ideas?

src/ui/dialogs/avatarupload.ts
mightymop: okay...

src/ui/dialogs/avatarupload.ts
mightymop: removed

src/ui/dialogs/avatarupload.ts
mightymop: We also use FileReader in Attachement.ts in getDataFromFile()

mightymop: but i see there is a generateThumbnail method we could reuse as util method

src/ui/dialogs/avatarupload.ts
mightymop: avatarupload.ts refactored...

src/util/SHA1Hash.ts
mightymop: renamed

src/ui/dialogs/avatarupload.ts
mightymop: avatarupload.ts refactored...

src/ui/dialogs/avatarupload.ts
mightymop: avatarupload.ts refactored...

src/ui/dialogs/avatarupload.ts
mightymop: avatarupload.ts refactored...

src/connection/services/Vcard.ts
mightymop: loadVcard calls parseVcard which will call parseVcardChildren and does not return a stanza in which we could merge the photo

src/connection/services/Vcard.ts
mightymop: problem is that we dont now if there is a additional method to shrink the images on server... so the photo will become a new hash. so i reload the vCard with the saved image and then i take this to send the presence hash

sualko:
Please check this pr again. There seams to be a merge error or something.

mightymop:
> 
> 
> Please check this pr again. There seams to be a merge error or something.

fixed it

src/connection/AbstractConnection.ts
```diff
@@ -206,9 +210,25 @@ abstract class AbstractConnection {
       //    presenceStanza.c('priority').t(priority[status]).up();
       // }
 
-      Log.debug('Send presence', presenceStanza.toString());
 
-      this.send(presenceStanza);
+          let hash = calculateHash(avatar);
```
sualko, line 214: `avatar` is a class, so this should fail.

PR #1000, opened 2021-01-13
feat: add muc user list on joindialog

mightymop:
... on hover over a search icon when group is not empty

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -427,7 +427,8 @@
     "no_results": "No results were found.",
     "trusted_domains": "Trusted domains",
     "one_domain_per_line": "Theese are domains from which attachments will be loaded automatically. (One domain per line)",
-    "setting-command-enable": "Allows to control different settings via text commands."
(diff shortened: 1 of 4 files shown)


sualko:
Congratulation! You opened the :one::zero::zero::zero:th ticket. :tada:

sualko:
I think the more appropriate place for the member list would be the next screen which shows also all other room information. So can you move your code to the `continueHandler`?

mightymop:
> 
> 
> I think the more appropriate place for the member list would be the next screen which shows also all other room information. So can you move your code to the `continueHandler`?

yes of course..

mightymop:
> Congratulation! You opened the 1️⃣0️⃣0️⃣0️⃣th ticket. 🎉

🍻

sualko:
> i only have to think abou what to do :D

Here you go: https://github.com/jsxc/jsxc/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement

PR #1008, opened 2021-01-17
feat: add XEP-0308 Last Message Correction

### LINKED ISSUE #746: XEP-0308: Last Message Correction
[XEP-0308](https://xmpp.org/extensions/xep-0308.html) defines a method for indicating that a message is a correction of the last sent message.

mightymop:
fix #746 
- [x] Receiving Part
- [x] Sending Part 

- [x] Changes after review

Long press / click on an outgoing message to edit the message and then simply press enter to send...

tested with converse.js as opponent and runs fine...

a1
a2

### DIFF
--- a/scss/partials/_window.scss
+++ b/scss/partials/_window.scss
@@ -323,6 +323,15 @@
     user-select: none;
     white-space: nowrap;
 
+    .jsxc-replace-icon {
+       background-image: url('../images/icons/edit.svg');
+       background-size: contain;
+       background-repeat: no-repeat;
+       background-position: center;
+       height: 15px;
+       width: 15px;
+    }
+
     .jsxc-timestamp {
       margin-right: 0.3em;
     }
--- a/scss/partials/_window.scss
+++ b/scss/partials/_window.scss
@@ -324,12 +324,12 @@
     white-space: nowrap;
 
     .jsxc-replace-icon {
-       background-image: url('../images/icons/edit.svg');
-       background-size: contain;
-       background-repeat: no-repeat;
-       background-position: center;
-       height: 15px;
-       width: 15px;
+      background-image: url("../images/icons/edit.svg");
+      background-position: center;
+      background-repeat: no-repeat;
+      background-size: contain;
+      height: 15px;
+      width: 15px;
     }
 
     .jsxc-timestamp {
--- a/src/ContactManager.ts
+++ b/src/ContactManager.ts
@@ -91,6 +91,10 @@ export default class ContactManager {
       storage.setItem('contacts', 'loaded', true);
    }
 
+   public getContacts(): { [key: string]: IContact } {
+      return this.contacts;
+   }
+
    public async add(contact: IContact): Promise<void> {
       this.contacts[contact.getId()] = contact;
 
--- a/src/Message.interface.ts
+++ b/src/Message.interface.ts
@@ -20,7 +20,6 @@ export enum MessageMark {
 }
 
 export interface IMessagePayload {
-
    peer: JID;
    direction: DIRECTION;
    attrId?: string;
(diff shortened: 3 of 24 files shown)


sualko (changes requested):
I think there is still some work to do:

- there are some points not implemented from the business rules and security considerations
- it will not work if attachment changes
- it will not work in muc
- you are not able to see the change log of a message

I would suggest to implement the replacements as linked list. This could be implemented with a `replacedBy` and `isCorrection` field (or something like that).

mightymop:
> 
> 
> I think there is still some work to do:
> 
>     * there are some points not implemented from the business rules and security considerations

This could be hard to implement... to we track the presences somewhere?

When used in a Multi-User Chat (XEP-0045) [3] context, corrections must not be allowed (by the receiver) for messages received before the sender joined the room - particularly a full JID leaving the room then rejoining and correcting a message SHOULD be disallowed, as the entity behind the full JID in the MUC may have changed.

>     * it will not work if attachment changes
> 
>     * it will not work in muc
> 
>     * you are not able to see the change log of a message
> 
> 
> I would suggest to implement the replacements as linked list. This could be implemented with a `replacedBy` and `isCorrection` field (or something like that).

sualko:
> Hmm, i would showing the edit icon with hover to show the correction log is enough...

It's ok for the beginning, but it would be nice to show at least the date of change.

> This could be hard to implement... to we track the presences somewhere?
> But i dont think the full jid will change on leaving and rejoining the room...
> I would add a check if old message and replacementmessage have the same fulljid...

The issue in a muc is that it depends on the kind of muc. For example in a public muc a user could leave and another join with the same nickname. Depending on the configuration we are not even aware of the users jid. (Note to self: We should not display the same avatar if the full jid is not known). We should probably add a temporary id for members that changes on every join. This way we could relate the message. This would also help with the avatar issue.

It's maybe also a good idea to allow only an edit of the latest message of a user (Conversations is doing the same and this is also coherent with the xep). 

> We could ignore replacements for security labels?

We don't support security labels, so I think we don't have to care about them.

> The Sender MUST NOT include a correction for a message with non-messaging payloads. For example, a sender MUST NOT include a correction for a roster item exchange request or a file transfer part.

But what happens if the user is adding an image url to the message? The http upload plugin will parse the url as attachment and therefore the url is missing in the corrected message. Therefore we should care about attachments.

> I guess we only need to log the original message ...

I'm not sure. For example you can't handle message marker correctly. If you change a message you should also reset the message marker to transferred, because otherwise the user maybe thinks that the contact has already read his update.

Another point is also mam. Should we support message correction in mam?

mightymop:
ok i need some help the the attachements and with muc if we dont have the real jid

mightymop:
> Another point is also mam. Should we support message correction in mam?

I guess it would make sence to integrate it too....

sualko (changes requested):
Honestly I'm not able to merge this, because...

- it's only one commit with 3 huge changes, this makes it impossible to review the whole thing in an acceptable time range.
- XEP-0421 (Anonymous unique occupant identifiers for MUCs) is not always available and I see no workaround for those cases.
- no code style.

If you have a branch with separate commits, please try to make meaningful commits, so that it's easier to review.

mightymop:
> Honestly I'm not able to merge this, because...
> 
> * it's only one commit with 3 huge changes, this makes it impossible to review the whole thing in an acceptable time range.
> * XEP-0421 (Anonymous unique occupant identifiers for MUCs) is not always available and I see no workaround for those cases.
> * no code style.
> 
> If you have a branch with separate commits, please try to make meaningful commits, so that it's easier to review.

:/ Bad news i only have this branch so no branch with separate commits available...

mightymop:
I have fixed the code styles via "yarn fix" and cherry-picked two of your commits to pass the pre-commit hook validation

sualko:
I'm currently developing a plugin for last message correction which solves the issue in a more isolated way. Until I'm done, I leave this open to copy and paste parts :-).

mightymop:
> I'm currently developing a plugin for last message correction which solves the issue in a more isolated way. Until I'm done, I leave this open to copy and paste parts :-).

Nice :)

In this way the XEP-424 would be a good addon too because xep308 and xep424 have a lot of common or dependent code

PR #1011, opened 2021-01-18
fix: Chatmarker should take care on type of message

### LINKED ISSUE #1007: Chatmarker should take care on type of message
### Expected behavior
If a message comes back with an type "error" the chatmarker should not show that the message was delivered
### Actual behavior
If i send a message to a user it could not be delivered the chatmarker shows received.
Plz have a look at the following xmpp stanza...

### Environment
- **JSXC version:**  master
- **Browser vendor and version:** latest chrome/ff x64 windoof
- **Any browser plugins enabled?** no
- **XMPP server vendor and version:** latest OF  
- **Is your XMPP server working with other clients as expected?** yes

#### Javascript

```

```

#### JSXC

```
```

#### XMPP

```
<message xmlns="jabber:client" id="9f072237-8a12-43e9-bdd0-4d37759535a8" to="test@MYDOMAIN.LOCAL/jsxc-913bb311" type="error" from="mops@MYDOMAIN.LOCAL">
   <body>AAAAAAAAAA</body>
   <origin-id xmlns="urn:xmpp:sid:0" id="9f072237-8a12-43e9-bdd0-4d37759535a8"/>
   <request xmlns="urn:xmpp:receipts"/>
   <active xmlns="http://jabber.org/protocol/chatstates"/>
   <markable xmlns="urn:xmpp:chat-markers:0"/>
   <error code="503" type="cancel">
      <service-unavailable xmlns="urn:ietf:params:xml:ns:xmpp-stanzas"/>
   </error>
</message>

<message xmlns="jabber:client" to="test@MYDOMAIN.LOCAL/jsxc-913bb311" type="error" from="mops@MYDOMAIN.LOCAL">
   <received xmlns="urn:xmpp:chat-markers:0" id="9f072237-8a12-43e9-bdd0-4d37759535a8"/>
   <store xmlns="urn:xmpp:hints"/>
   <error code="503" type="cancel">
      <service-unavailable xmlns="urn:ietf:params:xml:ns:xmpp-stanzas"/>
   </error>
</message>
```

#### Host

```
```

mightymop:
- if message type = error no chatmarker should be sent / shown in UI
- increased the character space between the two checkmarks so the second one is more visible...

fix #1007

### DIFF
--- a/scss/partials/_window.scss
+++ b/scss/partials/_window.scss
@@ -566,7 +566,7 @@
     color: $chatmessage-received;
     content: "✓✓";
     font-size: 12px;
-    letter-spacing: -0.5em;
+    letter-spacing: -0.3em;
     line-height: 12px;
     padding-right: 0.5em;
   }
--- a/src/plugins/chatMarkers/ChatMarkersPlugin.ts
+++ b/src/plugins/chatMarkers/ChatMarkersPlugin.ts
@@ -190,7 +190,11 @@ export default class ChatMarkersPlugin extends AbstractPlugin {
       let idAttr = messageElement.attr(ID);
       let fromAttr = messageElement.attr(FROM);
       let toAttr = messageElement.attr('to');
-      let typeAttr = stanzaElement.attr('type');
+      let typeAttr = messageElement.attr('type');
+
+      if (typeAttr === Message.MSGTYPE.GROUPCHAT || typeAttr === 'error') {
+         return true;
+      }
 
       let markableMessageId = markerElement.attr(ID);
       let marker = markerElement.prop('tagName').toLowerCase() as string;
@@ -208,7 +212,7 @@ export default class ChatMarkersPlugin extends AbstractPlugin {
             return true;
          }
 
-         if ($(stanza).attr('type') !== Message.MSGTYPE.GROUPCHAT && !isCarbon && !isMam) {
+         if (!isCarbon && !isMam) {
             let peer = new JID(fromAttr);
 
             this.sendReceived(idAttr, peer);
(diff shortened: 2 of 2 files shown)


mightymop:
Furthermore i realized that if a following message was sent successfully, the failed message will also be marked as sent... 
I think we should discuss this...

sualko:
I changed the code a bit, so it's using the message type and not the stanza type. They should be the same, but you never no.

> Furthermore i realized that if a following message was sent successfully, the failed message will also be marked as sent / delivered...

I added a check for that too. Can you test it?

sualko (approved):

mightymop:
> 
> 
> I changed the code a bit, so it's using the message type and not the stanza type. They should be the same, but you never no.
 
This works ... while the message could not be delivered no additional markers where shown, only the first in grey...
maybe we ad the error marker (the yellow triangle) on errors here?

> > Furthermore i realized that if a following message was sent successfully, the failed message will also be marked as sent / delivered...
> 
> I added a check for that too. Can you test it?

hm does not work... all messages which were send as the opponent were blocked got a mark and eye too after unblock und sending a regular message which was received and displayed...
I can see in the console that i only get the marker with id from the message that i had send after i unblocked the user again... 
So there is still this error ...

sualko:
Do those messages show the warning icon on the top right corner? I think we don't handle message error probably on a global level.

mightymop:
> Do those messages show the warning icon on the top right corner? I think we don't handle message error probably on a global level.

No they do not

sualko:
> No they do not

If they are blocked from blocking command there should be a warning, maybe OF is not including the correct marker due to there implementation with privacy lists.

>  I think we don't handle message error probably on a global level.

We should probably tackle that sometime soon. Not sure what the best approach would be, but my first guess would be a error class where plugins could register different error cases / callbacks, so that you have only one global message error handler. Maybe we can also use a pipe for that.

mightymop:
> > No they do not
> 
> If they are blocked from blocking command there should be a warning, maybe OF is not including the correct marker due to there implementation with privacy lists.
> 
Thats why i asked for the other error tags in the onMessageError method of the BlockingPlugin... 

> > I think we don't handle message error probably on a global level.
> 
> We should probably tackle that sometime soon. Not sure what the best approach would be, but my first guess would be a error class where plugins could register different error cases / callbacks, so that you have only one global message error handler. Maybe we can also use a pipe for that.

Sounds like a goog idea for the future... but should be implemented in another PR

sualko:
Since the original error is solved, I merge this now.

PR #1013, opened 2021-01-18
catch error messages

sualko:
We should probably adapt the error message to the different error states defined in https://tools.ietf.org/html/rfc6120#section-8.3.

cc @mightymop


mightymop:
As for the chatmarkers it works. So the chatmarkers for error message were not set to received... etc. 
But what about the yellow triangel in the upper right corner you mentioned? 
All in all it looks good.

sualko:
> But what about the yellow triangel in the upper right corner you mentioned?

If there is a error for a message, there should be an icon:

image

mightymop:
I dont get the Icons:/ i have to debug this to know what causes that behavoir

sualko:
> furthermore contact is sender because it came back as error i think... (but this is also a bug in OF -.- because to and from MUST be switched in error message)

As I mentioned earlier, OpenFire is non standard conform XMPP server. If we add workarounds for every bug the code will be unreadable and it's likely that it will introduce issues with standard conform implementations.

> but i think if we would get the blocked tag the message is wrong then.. because it should be "You were blocked by this jid" but now it is "You have blocked this JID." (So actually it is the other way around).

According to [XEP-0191](https://xmpp.org/extensions/xep-0191.html) our implementation should be correct. If you send a message to a jid you blocked, the server should response with a blocked tag. If the blocked jid is sending a message, he should get an service-unavailable. Maybe OF is messing this up.

> but arent we able to get the from by the message id?

This feels like an dangerous operation. The attr id only needs to be unique per session. So the server could send us an own error message with the same id. 

1bd951f is only working if the chat window is open. If we think about s2s messages, the error could be delayed and the window closed. As mentioned earlier I think we should rather fix OF, instead of JSXC.

mightymop:
> 
> 
> > furthermore contact is sender because it came back as error i think... (but this is also a bug in OF -.- because to and from MUST be switched in error message)
> 
> As I mentioned earlier, OpenFire is non standard conform XMPP server. If we add workarounds for every bug the code will be unreadable and it's likely that it will introduce issues with standard conform implementations.
> 
> > but i think if we would get the blocked tag the message is wrong then.. because it should be "You were blocked by this jid" but now it is "You have blocked this JID." (So actually it is the other way around).
> 
> According to [XEP-0191](https://xmpp.org/extensions/xep-0191.html) our implementation should be correct. If you send a message to a jid you blocked, the server should response with a blocked tag. If the blocked jid is sending a message, he should get an service-unavailable. Maybe OF is messing this up.
> 
> > but arent we able to get the from by the message id?
> 
> This feels like an dangerous operation. The attr id only needs to be unique per session. So the server could send us an own error message with the same id.
> 
> [1bd951f](https://github.com/jsxc/jsxc/commit/1bd951f759ce77c67f69948761cb86f75cb9f418) is only working if the chat window is open. If we think about s2s messages, the error could be delayed and the window closed. As mentioned earlier I think we should rather fix OF, instead of JSXC.

guusdk from igniterealtime is working on the fixes for OF... i sent him all my test results too to fix it...

Furthermore you are right... [1bd951f](https://github.com/jsxc/jsxc/commit/1bd951f759ce77c67f69948761cb86f75cb9f418) only works for opened windows and my intend was it to use it only as fallback, because without it the messages were dropped... so with this fallback we get a little chance to show the yellow error triangle somehow

sualko:
> guusdk from igniterealtime is working on the fixes for OF... i sent him all my test results too to fix it...

That's nice to hear.

> Furthermore you are right... 1bd951f only works for opened windows

Not any more. Since we use the same id for attribute id and stanza id we can directly get the message from the storage. As second fallback we are using your approach.

sualko:
I thought it's maybe nice to show the error message instead the need to hover the message. This is also useful on touch devices. What do you think?

image

mightymop:
4k

> I thought it's maybe nice to show the error message instead the need to hover the message. This is also useful on touch devices. What do you think?
> 
> image

Nice ... 👍 
We could also use a combination of both... first showing the the triangle and with click it presenting your text for 5 seconds under the message... so we would save much space on small screens....

PR #1018, opened 2021-02-24
fix: message presend process not working

anvyzhang:
Problem is

* When I process a message in presend functions, like do some string replacement on message content.
* Send the message out, sender's message area shows a new message without those string replacement .
* But the receiver got a new message which is processed and string replaced.

This PR can fix this problem

### DIFF
--- a/src/ui/ChatWindow.ts
+++ b/src/ui/ChatWindow.ts
@@ -463,13 +463,12 @@ export default class ChatWindow {
          unread: false,
       });
 
-      this.getTranscript().pushMessage(message);
-
       this.clearAttachment();
 
       let pipe = this.getAccount().getPipe('preSendMessage');
 
       pipe.run(this.contact, message).then(([contact, message]) => {
+         this.getTranscript().pushMessage(message);
          this.getAccount().getConnection().sendMessage(message);
       }).catch(err => {
          Log.warn('Error during preSendMessage pipe', err);


src/ui/ChatWindow.ts
```diff
@@ -474,7 +472,9 @@ export default class ChatWindow {
       }).catch(err => {
          Log.warn('Error during preSendMessage pipe', err);
       });
-
+      
+      this.getTranscript().pushMessage(message);
```
sualko, line 476: At this point it is not guaranteed that the pipe has finished (e.g. a plugin could wait for a http request to finish).

sualko (changes requested):
To solve the issue the message should be added after the pipe promise resolves, or we the [chat message ui](https://github.com/jsxc/jsxc/blob/master/src/ui/ChatWindowMessage.ts#L217) should listen for text updates. I would prefer the second approach, because otherwise the user has to wait sometime until the message is shown in the chat area.

sualko:
@anvyzhang no need to close this pr. It's a good reminder to fix the issue and if you have time you can also force push your fix for that.

anvyzhang:
Ok, but I think it's better to wait some time(short enouth, otherwise mabe we should find another better way to implement our idea rather than use presend functions), because sync message process may made some splash or refresh on message, which is harmful for endusers.

sualko:
If you could tell me what you try to do, I maybe can help you.

anvyzhang:
I made the message more rich formatted, to better display images, links, and even robot message card style. @sualko

sualko:
> I made the message more rich formatted, to better display images, links, and even robot message card style.

For this you can register a message formatter via `Message.addFormatter`. Since I think that will fix your issue, I close this pr. If not, please provide an example.

PR #1021, opened 2021-03-25
XEP84 refactored

mightymop:
-I did my best, but found out that when both avatar plugins are enabled there is some trouble when uploading a new / or deleting an avatar.

the parameter "avatar" in publishAvatarProcessor will become undefined although i put a defined one at pipe call

### DIFF
--- a/src/plugins/AvatarPEPPlugin.ts
+++ b/src/plugins/AvatarPEPPlugin.ts
@@ -110,7 +110,7 @@ export default class AvatarPEPPlugin extends AbstractPlugin {
    private onMessageAvatarUpdate = stanza => {
       let from = new JID($(stanza).attr('from'));
       let metadata = $(stanza).find('metadata[xmlns="urn:xmpp:avatar:metadata"]');
-  
+
       if (metadata.length > 0) {
          let info = metadata.find('info');
          let contact = this.pluginAPI.getContact(from);
@@ -192,7 +192,7 @@ export default class AvatarPEPPlugin extends AbstractPlugin {
                   let hash = $(info).attr('id');
                   if (hash && hash.length > 0) {
                      let typeval = info.attr('type');
-                     
+
                      let regextypeval = new RegExp(/image\/(\*|png|jpg|jpeg)/igm);
                      if (!regextypeval.test(typeval))
                      {
--- a/src/plugins/AvatarPEPPlugin.ts
+++ b/src/plugins/AvatarPEPPlugin.ts
@@ -70,6 +70,20 @@ export default class AvatarPEPPlugin extends AbstractPlugin {
       } else {
          let data = $build('data', { xmlns: UANS_DATA }).t(avatar.getData()).tree();
 
+         const b64str = avatar.getData().replace(/^.+;base64,/, '');
+
+         if (b64str[0]!=='/' && //base64 mimeType for jpeg
+            b64str[0]!=='i') //base64 mimeType for png
+         {
+            throw new Error('Only jpeg and png files are supported.');
+         }
+
+         let regextypeval = new RegExp(/^(?:[A-Za-z0-9+\/]{4})*(?:[A-Za-z0-9+\/]{2}==|[A-Za-z0-9+\/]{3}=)?$/igm);
+         if (!regextypeval.test(b64str))
+         {
+            throw new Error('Data Source is not a valid base64 string!');
+         }
+
          return connection
             .getPEPService()
             .publish(UANS_DATA, data, UANS_DATA)
@@ -87,11 +101,7 @@ export default class AvatarPEPPlugin extends AbstractPlugin {
                     .getPEPService()
                     .publish(UANS_METADATA, metadata, UANS_METADATA)
                     .then(function (result) {
-                        if ($(result).attr('type') === 'result') {
-                            return [undefined];
-                        } else {
-                            return [avatar];
-                        }
+                        return [avatar];
                      });
             });
       }
--- a/src/plugins/AvatarPEPPlugin.ts
+++ b/src/plugins/AvatarPEPPlugin.ts
@@ -196,16 +196,35 @@ export default class AvatarPEPPlugin extends AbstractPlugin {
                   let hash = $(info).attr('id');
                   if (hash && hash.length > 0) {
                      let typeval = info.attr('type');
+                     
+                     let regextypeval = new RegExp(/image\/(\*|png|jpg|jpeg)/igm);
+                     if (!regextypeval.test(typeval))
+                     {
+                        throw new Error('Mimetype not allowed');
+                     }
 
                      return connection
                         .getPEPService()
                         .retrieveItems(UANS_DATA, contact.getJid().bare)
                         .then(data => {
                            if (data && $(data).text() && $(data).text().trim().length > 0) {
-                              let src = $(data)
+                              let src= $(data)
                                  .text()
-                                 .replace(/[\t\r\n\f]/gi, '');
+                                 .replace(/[\t\r\n\f]/gim, '');
                               const b64str = src.replace(/^.+;base64,/, '');
+
+                              if (b64str[0]!=='/' && //base64 mimeType for jpeg
+                                  b64str[0]!=='i') //base64 mimeType for png
+                              {
+                                 throw new Error('Only jpeg and png files are supported.');
+                              }
+
+                              let regextypeval = new RegExp(/^(?:[A-Za-z0-9+\/]{4})*(?:[A-Za-z0-9+\/]{2}==|[A-Za-z0-9+\/]{3}=)?$/igm);
+                              if (!regextypeval.test(b64str))
+                              {
+                                 throw new Error('Data Source is not a valid base64 string!');
+                              }
+
                               this.getStorage().setItem(contact.getJid().bare, Hash.SHA1FromBase64(b64str));
                               return { src: 'data:' + typeval + ';base64,' + src, type: typeval };
                            } else {
--- a/src/plugins/AvatarPEPPlugin.ts
+++ b/src/plugins/AvatarPEPPlugin.ts
@@ -1,0 +1,245 @@
+import { AbstractPlugin, IMetaData } from '../plugin/AbstractPlugin';
+import PluginAPI from '../plugin/PluginAPI';
+import { IContact } from '../Contact.interface';
+import Avatar from '../Avatar';
+import JID from '../JID';
+import Translation from '@util/Translation';
+import Hash from '@util/Hash';
+import { IAvatar } from '@src/Avatar.interface';
+import AvatarUI from '../ui/AvatarSet';
+import FileHelper from '@util/FileHelper';
+
+const MIN_VERSION = '4.0.0';
+const MAX_VERSION = '99.0.0';
+
+const UANS_BASE = 'urn:xmpp:avatar';
+const UANS_METADATA = +UANS_BASE + ':metadata';
+const UANS_NOTIFY = UANS_METADATA + '+notify';
+const UANS_DATA = UANS_BASE + ':data';
+
+export default class AvatarPEPPlugin extends AbstractPlugin {
+   public static getId(): string {
+      return 'pep-avatars';
+   }
+
+   public static getName(): string {
+      return 'PEP-based Avatars';
+   }
+
+   public static getMetaData(): IMetaData {
+      return {
+         description: Translation.t('setting-pep-avatar-enable'),
+         xeps: [
+            {
+               id: 'XEP-0084',
+               name: 'User Avatar',
+               version: '1.1.4',
+            },
+         ],
+      };
+   }
+
+   constructor(pluginAPI: PluginAPI) {
+      super(MIN_VERSION, MAX_VERSION, pluginAPI);
+
+      let connection = pluginAPI.getConnection();
+
+      connection.registerHandler(this.onMessageAvatarUpdate, 'http://jabber.org/protocol/pubsub#event', 'message');
+
+      pluginAPI.addAvatarProcessor(this.avatarProcessor, 49);
+      pluginAPI.addPublishAvatarProcessor(this.publishAvatarProcessor, 49);
+      pluginAPI.addFeature(UANS_NOTIFY);
+   }
+
+   public getStorage() {
+      return this.pluginAPI.getStorage();
+   }
+
+   private publishAvatarProcessor = (avatar: IAvatar | null): Promise<[IAvatar]> => {
+      let connection = this.pluginAPI.getConnection();
+
+      if (!avatar || avatar.getData() === undefined) {
+         let item = $build('metadata', { xmlns: UANS_METADATA });
+
+         return connection
+            .getPEPService()
+            .publish(UANS_METADATA, item.tree(), UANS_METADATA)
+            .then(function (result) {
+               if ($(result).attr('type') === 'result') {
+                  return [undefined];
+               } else {
+                  return [avatar];
+               }
+            });
+      } else {
+         let data = $build('data', { xmlns: UANS_DATA }).t(avatar.getData()).tree();
+
+         return connection
+            .getPEPService()
+            .publish(UANS_DATA, data, UANS_DATA)
+            .then((result: any) => {
+               if ($(result).attr('type') === 'result') {
+                  let metadata;
+                  let hash = Hash.SHA1FromBase64(avatar.getData());
+                  let i = new Image();
+                  let iheight = 0;
+                  let iwidth = 0;
+                  let size = FileHelper.getFileSizeFromBase64(avatar.getData());
+
+                  i.onload = function () {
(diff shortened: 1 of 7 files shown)


src/ui/AvatarSet.ts
```diff
@@ -28,6 +27,15 @@ export default class AvatarSet {
       this.reload();
    }
 
+   public update(src) {
```
sualko, line 30: This function should not be needed, because reload is requesting the avatar and updating all elements.

sualko:
Thanks for taking the time to refactor your changes. I will review it next week.

src/ui/AvatarSet.ts
mightymop: This function is called by the messageevent which allready has the imagedata in it ... i wanted to prevent to much network traffic by reloading again

sualko: It should not produce any additional network traffic, if you store the avatar at that point. For this reason I removed the function.

src/plugins/AvatarPEPPlugin.ts
```diff
@@ -0,0 +1,245 @@
...
+   constructor(pluginAPI: PluginAPI) {
+      super(MIN_VERSION, MAX_VERSION, pluginAPI);
+
+      let connection = pluginAPI.getConnection();
+
+      connection.registerHandler(this.onMessageAvatarUpdate, 'http://jabber.org/protocol/pubsub#event', 'message');
+
+      pluginAPI.addAvatarProcessor(this.avatarProcessor, 49);
+      pluginAPI.addPublishAvatarProcessor(this.publishAvatarProcessor, 49);
+      pluginAPI.addFeature(UANS_NOTIFY);
```
sualko, line 51: Should use `connection.getPEPService().subscribe()` which makes the handler above obsolete.

src/plugins/AvatarPEPPlugin.ts
```diff
@@ -0,0 +1,245 @@
...
+   }
+
+   private publishAvatarProcessor = (avatar: IAvatar | null): Promise<[IAvatar]> => {
+      let connection = this.pluginAPI.getConnection();
+
+      if (!avatar || avatar.getData() === undefined) {
+         let item = $build('metadata', { xmlns: UANS_METADATA });
+
+         return connection
+            .getPEPService()
```
sualko, line 65: can go to the top

src/plugins/AvatarPEPPlugin.ts
```diff
@@ -0,0 +1,245 @@
...
+            .publish(UANS_METADATA, item.tree(), UANS_METADATA)
+            .then(function (result) {
+               if ($(result).attr('type') === 'result') {
+                  return [undefined];
+               } else {
+                  return [avatar];
+               }
+            });
+      } else {
```
sualko, line 74: Since the if block returns, we don't need the else.

src/plugins/AvatarPEPPlugin.ts
```diff
@@ -0,0 +1,245 @@
...
+         let item = $build('metadata', { xmlns: UANS_METADATA });
+
+         return connection
+            .getPEPService()
+            .publish(UANS_METADATA, item.tree(), UANS_METADATA)
+            .then(function (result) {
+               if ($(result).attr('type') === 'result') {
+                  return [undefined];
+               } else {
+                  return [avatar];
```
sualko, line 71: Since there should be no avatar at this point, the if doesn't have any impact. Publish also resolves the promise only if the request was successful, therefore the type check is useless.

src/plugins/AvatarPEPPlugin.ts
```diff
@@ -0,0 +1,245 @@
...
+               }
+            });
+      } else {
+         let data = $build('data', { xmlns: UANS_DATA }).t(avatar.getData()).tree();
+
+         return connection
+            .getPEPService()
+            .publish(UANS_DATA, data, UANS_DATA)
+            .then((result: any) => {
```
sualko, line 80: We can use async/await here.

src/plugins/AvatarPEPPlugin.ts
```diff
@@ -0,0 +1,245 @@
...
+               if ($(result).attr('type') === 'result') {
```
sualko, line 81: As I wrote above, this check is not required.

src/plugins/AvatarPEPPlugin.ts
```diff
@@ -0,0 +1,245 @@
...
+                  let metadata;
```
sualko, line 82: No need to declare it here. If you declare it where you need it, it also helps ts to determine the type.

src/plugins/AvatarPEPPlugin.ts
```diff
@@ -0,0 +1,245 @@
...
+                  let hash = Hash.SHA1FromBase64(avatar.getData());
+                  let i = new Image();
+                  let iheight = 0;
+                  let iwidth = 0;
+                  let size = FileHelper.getFileSizeFromBase64(avatar.getData());
+
+                  i.onload = function () {
+                     iheight = i.height;
+                     iwidth = i.width;
```
sualko, line 91: Onload is asynchron, therefore the height and width could be 0. Since those attributes are not require, we don't have to provide them.

src/plugins/AvatarPEPPlugin.ts
```diff
@@ -0,0 +1,245 @@
...
+                  };
+                  i.src = 'data:' + avatar.getType() + ';base64,' + avatar.getData();
+                  metadata = $build('metadata', { xmlns: UANS_METADATA })
+                     .c('info', { bytes: size, id: hash, height: iheight, width: iwidth, type: avatar.getType() })
```
sualko, line 95: Do we verify that we have only a png file?

src/plugins/AvatarPEPPlugin.ts
```diff
@@ -0,0 +1,245 @@
...
+                     .tree();
+
+                  return connection
+                     .getPEPService()
+                     .publish(UANS_METADATA, metadata, UANS_METADATA)
+                     .then(function (result) {
+                        if ($(result).attr('type') === 'result') {
+                           return [undefined];
```
sualko, line 103: This would prevent other plugins to update the avatar as well.

src/plugins/AvatarPEPPlugin.ts
```diff
@@ -0,0 +1,245 @@
...
+               let avatarUI = AvatarUI.get(contact);
+               this.getStorage().setItem(contact.getJid().bare, hash);
+               avatarUI.reload();
+            }
+         } else {
+            let avatarUI = AvatarUI.get(contact);
+            this.getStorage().setItem(contact.getJid().bare, '');
+            avatarUI.reload();
+         }
+      } else if (data.length > 0) {
```
sualko, line 142: Since we subscribe only to metadata updates, this block should not be needed.

src/plugins/AvatarPEPPlugin.ts
```diff
@@ -0,0 +1,245 @@
...
+         if (!contact) {
+            this.pluginAPI.Log.warn('No contact found for', from);
+            return true;
+         }
+
+         if (info.length > 0) {
+            let hash = $(info).attr('id');
+
+            let storedHash = this.getStorage().getItem(from.bare);
+            if (storedHash === undefined || hash !== storedHash) {
```
sualko, line 132: We should also check if the hash is valid.

src/plugins/AvatarPEPPlugin.ts
```diff
@@ -0,0 +1,245 @@
...
+         avatar = new Avatar(hash);
+      } catch (err) {
+         try {
+            const avatarObject = await this.getAvatar(contact);
+
+            if (avatarObject) {
+               avatar = new Avatar(hash, avatarObject.type, avatarObject.src);
+               return [contact, avatar];
+            } else {
+               this.pluginAPI.Log.warn('No local cached avatar found');
```
sualko, line 193: `this.getAvatar` should retrieve the avatar, so the message is wrong.

src/plugins/AvatarPEPPlugin.ts
```diff
@@ -0,0 +1,245 @@
...
+
+      if (!hash || avatar) {
+         return [contact, avatar];
+      }
+
+      try {
+         avatar = new Avatar(hash);
+      } catch (err) {
+         try {
+            const avatarObject = await this.getAvatar(contact);
```
sualko, line 187: If there is no avatar, `this.getAvatar` should trhow an error, so the following check is not needed. Makes the code a bit easier to read.

src/plugins/AvatarPEPPlugin.ts
```diff
@@ -0,0 +1,245 @@
...
+      return [contact, avatar];
+   };
+
+   private async getAvatar(contact: IContact): Promise<{ src: string; type: string }> {
+      let connection = this.pluginAPI.getConnection();
+
+      return connection
+         .getPEPService()
+         .retrieveItems(UANS_METADATA, contact.getJid().bare)
+         .then(meta => {
```
sualko, line 211: Please use async/await.

src/plugins/AvatarPEPPlugin.ts
```diff
@@ -0,0 +1,245 @@
...
+
+      return [contact, avatar];
+   };
+
+   private async getAvatar(contact: IContact): Promise<{ src: string; type: string }> {
+      let connection = this.pluginAPI.getConnection();
+
+      return connection
+         .getPEPService()
+         .retrieveItems(UANS_METADATA, contact.getJid().bare)
```
sualko, line 210: I would add a `hash` parameter, because sometimes we have the hash already from the metadata notification.

src/plugins/AvatarPEPPlugin.ts
```diff
@@ -0,0 +1,245 @@
...
+         .then(meta => {
+            let metadata = $(meta).find('metadata[xmlns="urn:xmpp:avatar:metadata"]');
+
+            if (metadata.length > 0) {
+               let info = metadata.find('info');
+
+               if (info && info.length > 0) {
+                  let hash = $(info).attr('id');
```
sualko, line 218: `info` is already a jquery element.

src/plugins/AvatarPEPPlugin.ts
```diff
@@ -0,0 +1,245 @@
...
+      return connection
+         .getPEPService()
+         .retrieveItems(UANS_METADATA, contact.getJid().bare)
+         .then(meta => {
+            let metadata = $(meta).find('metadata[xmlns="urn:xmpp:avatar:metadata"]');
+
+            if (metadata.length > 0) {
+               let info = metadata.find('info');
+
+               if (info && info.length > 0) {
```
sualko, line 217: Should be exactly one element.

src/plugins/AvatarPEPPlugin.ts
```diff
@@ -0,0 +1,245 @@
...
+                  let hash = $(info).attr('id');
+                  if (hash && hash.length > 0) {
+                     let typeval = $(info).attr('type');
+
+                     let result = connection
+                        .getPEPService()
+                        .retrieveItems(UANS_DATA, contact.getJid().bare)
+                        .then(data => {
+                           if (data && $(data).text() && $(data).text().trim().length > 0) {
```
sualko, line 226: We can have multiple data elements.

src/plugins/AvatarPEPPlugin.ts
```diff
@@ -0,0 +1,245 @@
...
+                              let src = $(data)
+                                 .text()
+                                 .replace(/[\t\r\n\f]/gi, '');
+                              const b64str = src.replace(/^.+;base64,/, '');
+                              this.getStorage().setItem(contact.getJid().bare, Hash.SHA1FromBase64(b64str));
+                              return { src: 'data:' + typeval + ';base64,' + src, type: typeval };
```
sualko, line 232: We need to check if type and src is valid.

sualko (changes requested):
I already made some smaller adjustments, but there is still some work to do. If you have time it would be great if you can have a look. Otherwise I will try to merge it next week.

mightymop:
> 
> 
> I already made some smaller adjustments, but there is still some work to do. If you have time it would be great if you can have a look. Otherwise I will try to merge it next week.

I am very busy with family and work at the moment so there is sadly no guarantee that i will have time this week :/

src/plugins/AvatarPEPPlugin.ts
mightymop: ok we could check if sha of src equals to known sha hash but checking the type? detecting mimetype via magic numbers?

src/plugins/AvatarPEPPlugin.ts
mightymop: Should we care of multiple data elements or just use the first one?

src/plugins/AvatarPEPPlugin.ts
mightymop: Doesnt the server sends pep data to user when he has succesfully logged in ? Then it would be useful

src/plugins/AvatarPEPPlugin.ts
mightymop: you mean that the user does not append/sideload other code into the image?

src/plugins/AvatarPEPPlugin.ts
mightymop: removed

src/plugins/AvatarPEPPlugin.ts
mightymop: removed check

src/plugins/AvatarPEPPlugin.ts
mightymop: removed

src/plugins/AvatarPEPPlugin.ts
mightymop: changed

src/plugins/AvatarPEPPlugin.ts
mightymop: removed

src/plugins/AvatarPEPPlugin.ts
mightymop: why?

src/plugins/AvatarPEPPlugin.ts
mightymop: removed $()

src/plugins/AvatarPEPPlugin.ts
mightymop: changed to ===1

mightymop:
i changed some of your code requirements... would you do the rest?

src/plugins/AvatarPEPPlugin.ts
sualko: The XEP requires us to publish at least a png (beside possible other formats). Do we verify at some point that the user uploaded a png, or do we convert the image somewhere?

src/plugins/AvatarPEPPlugin.ts
sualko: Because in the pipe every processor is called one after the other and the return value is always the "input" value of the next processor. Therefore the next processor will think that the avatar was deleted.

src/plugins/AvatarPEPPlugin.ts
sualko: No. The server should not send any data without request or subscription.

src/plugins/AvatarPEPPlugin.ts
sualko: I would pick the png file, but we can also define another order. Maybe it makes also sense to pick the one with the resolution nearest to 44px, or something.

src/plugins/AvatarPEPPlugin.ts
sualko: I only want to prevent XSS, so we should check if typeval is a valid mimetype and src contains only valid base64 characters.

mightymop: i have added some checks now:
check data source for jpeg/png file
check for valid base64 characters
check mimetype for jpeg or png

src/plugins/AvatarPEPPlugin.ts
mightymop: i added some checks and now only png or jpeg are supported...

src/plugins/AvatarPEPPlugin.ts
mightymop: so we should return avatar variable then..

mightymop: changed...

src/plugins/AvatarPEPPlugin.ts
mightymop: now also checking the uploaded data for valid base64 string and for jpeg or png

src/plugins/AvatarPEPPlugin.ts
mightymop: okay.. then i could delete it..

PR #1059, opened 2021-09-24
style: dont hide groups when using "only show online contacts"

mightymop:
If enabled the "only show online contacts" it should not hide the groups in the roster too.
I changed the css to not hide data-type=groupchat and removed the offline label in the li-element.
I think if someone dont want to see the groups, they could be collapsed.


sualko:
I still do not get the use case. Join = Online, so the behavior of the "only show online contacts" label makes sense to me.

If you have a good example, which helps to understand the scenario, please reopen this pr. I'm open for discussion.

PR #1060, opened 2021-09-27
feat: show "Load Button" if message area is empty

mightymop:
and show button, if chat history was cleared
fix: if chatwindow area was dragged bigger than the scoll did not show the button if top was reached...

### DIFF
--- a/src/plugins/mam/Plugin.ts
+++ b/src/plugins/mam/Plugin.ts
@@ -150,7 +150,11 @@ export default class MessageArchiveManagementPlugin extends AbstractPlugin {
       element.append(spanElement);
 
       messageAreaElement.on('scroll', function () {
-         if (messageAreaElement.height() + this.scrollTop < 42 && !archive.isExhausted()) {
+         let scrollTop : number     = messageAreaElement[0].scrollTop;
+         if (scrollTop<0)
+            scrollTop=scrollTop*(-1);
+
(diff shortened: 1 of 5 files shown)


src/plugins/mam/Plugin.ts
```diff
@@ -64,6 +64,8 @@ export default class MessageArchiveManagementPlugin extends AbstractPlugin {
          if (this.supportCache[archiveJid.bare]) {
             this.getArchive(contact.getJid()).clear();
          }
+
+         this.addLoadButtonIfEnabled(chatWindow, contact);
```
sualko, line 68: This will add a second button.

src/plugins/mam/Plugin.ts
```diff
@@ -150,7 +152,11 @@ export default class MessageArchiveManagementPlugin extends AbstractPlugin {
       element.append(spanElement);
 
       messageAreaElement.on('scroll', function () {
-         if (messageAreaElement.height() + this.scrollTop < 42 && !archive.isExhausted()) {
+         let scrollTop : number     = messageAreaElement[0].scrollTop;
+         if (scrollTop<0)
```
sualko, line 156: `scrollTop` is negative, because the flex layout is reverse.

sualko:
I changed the pr a bit, but should work as expected. Thanks again for your work.

PR #1062, opened 2021-11-09
fix(i18n): Add missing spaces to NL_nl

guusdk:
At various places where placeholders were used, a space was missing. This causes a small annoyance where text would should up likethis (sic).


mightymop (approved):
Although i dont speak this language, the PR seems to be good.

PR #1064, opened 2021-11-17
i18n nl-NL: various fixes and improvements

guusdk:


sualko:
Thanks again. If you like you can also contribute translations via https://webtranslateit.com/en/projects/10365-JSXC

Issue #1076, opened 2022-02-15
Nickname change leaves MUC room dialog unmodifiable


guusdk:
Trying to update Openfire's JSXC plugin to update to version 4.4.0, I'm noticing the following:

I can join (and participate in) MUC rooms just fine, up until I try to change my nickname (using `/nick newnickname`). The UI element that shows the MUC room chat turns grey (as if it's disabled), and I can no longer type messages into it. I keep receiving messages and events from other occupants.

I've seen this with a chatroom hosted on Openfire (the same server as what I was logged into using JSXC), as well as in chatrooms hosted on xmpp.org (which I joined through federation).

sualko:
Can you enable debug mode and attach your console output? Would be nice to see, what is happening.

guusdk:
This is the console log as saved by Chrome, that was generated in this environment:

- Build and start a new [Openfire 4.7.0 server](https://github.com/igniterealtime/Openfire/releases/tag/v4.7.0) in demoboot mode (which automatically provisions it with a couple of accounts) (`mvn clean package && ./distribution/target/distribution-base/bin/openfire.sh -debug`)
- Installed the Openfire plugin that bundles JSXC 4.4.0 using the Openfire admin console (based on this commit: https://github.com/guusdk/openfire-jsxc-plugin/commit/79eb47462d73cd71f207817e88b242c21b150d5c - build using `mvn clean package`, then rename `target/jsxc-openfire-plugin-assembly.jar` to `jsxc.jar` which can be uploaded to Openfire).
- Set the `jsxc.config.debug_enabled` property value to 'true' in the Openfire admin console, to cause JSXC's debug mode to be enabled.
- Open a new "incognito" Chrome window. Start the developer tools.
- Navigate to `http://example.org:7070/jsxc/`
- Login with 'john' / 'secret'
- Create a new chat room named 'testroom', accepting all defaults.
- type a message in the room: 'test'
- type `/nick DifferentNick`

This is now what my browser window looks like.

![Screenshot from 2022-02-15 13-54-52](https://user-images.githubusercontent.com/4253898/154066242-9f116ae0-db51-40e1-851b-91b191f0d72b.png)

Full console log output: [auth-createmuc-join-changenickname.log](https://github.com/jsxc/jsxc/files/8069163/auth-createmuc-join-changenickname.log)

guusdk:
I've got the same issue when using JSXC 4.3.1.

mightymop:
[...]

PR #1078, opened 2022-02-19
fix: Nickname change leaves MUC room dialog unmodifiable

mightymop:
fix: #1076


sualko:
Thanks for taking the time to investigate the issue. It enabled me to quickly find a better fix, because the shutdown is handled a few lines above your change and so we do not need to call shutdown in the rename member method.

PR #1087, opened 2022-02-26
feat: XEP-0424 Message Retraction

mightymop:
This Plugin is based on your LMC Plugin ;)


Neustradamus:
@sualko: What do you think?

PR #1089, opened 2022-02-26
Improve chatwindow

mightymop:
feat: formating tools
feat: add scroll down button when not at bottom
feat: dont scroll on new messages when manually scrolled up, resume autoscroll, when on bottom
feat: show message length on input
feat: blinking chatwindow title bar when chatwindow is minimized and a new message arrived
feat: if in input and use Key_Up jump to LMC edit mode
fix: css styles


Neustradamus:
@sualko: What do you think?

